PR #84, opened 2021-03-19
Potentially improve HashMap deserialization perf

matklad:
drive by change, didn't do any benchmarks, but seems like this can't hurt, right?


djkoloski:
Looks good to me, every little bit helps!

PR #106, opened 2021-04-15
Run tests in WebAssembly in CI

as-com:
I want to use rkyv in WebAssembly, so this PR makes the necessary changes to rkyv_test to make it compile and run on WASM targets. It also adds a step to CI to run these tests in a Node.js environment. 

Note that this does not (yet) ensure that serialized data are portable between x86 and WASM. I also had to disable rkyv_dyn in WASM tests due to its reliance on the inventory crate, [which does not support WASM targets](https://github.com/dtolnay/inventory/issues/13).

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -16,7 +16,14 @@ jobs:
 
     steps:
     - uses: actions/checkout@v2
+    - uses: actions-rs/install@v0.1
+      with:
+        crate: wasm-pack
+        version: latest
+        use-tool-cache: true
     - name: Build
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
+    - name: Run wasm tests
+      run: cd rkyv_test && wasm-pack test --node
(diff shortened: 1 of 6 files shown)


djkoloski:
Looks great, thanks for the help. I made a few small changes, mostly trying to push everything into a `wasm` feature. There's a [blocking bug](https://github.com/rustwasm/wasm-pack/issues/698) preventing this from happening completely, but it looks like `wasm-pack` is pushing toward a new release so that may change.

djkoloski:
You can compile the same source to x86 and WASM and send messages between processes. It should also work across platforms, but test coverage for this is not very comprehensive. For example, you may need to enable the `strict` feature, which imposes stricter representation requirements on archived structs. In the past, communication between x86 and WASM has been successful with `strict` enabled.

The note on this PR is that the test suite does not exercise sending messages serialized in one environment to the other, which would make the tests more comprehensive. An equivalent exercise would be serializing to a file in both environment and ensuring that the files are identical. If you have ideas on how to get these kinds of tests working in CI they would be greatly appreciated! 🙂

PR #139, opened 2021-05-21
Refine HashMap implementation

moelife-coder:
Recently, I was doing a research about massive-data management and I realized that the performance of my prototype is lower than expected. After some investigation, I figured that serializing `HashMap` is far more expensive than I expect. Therefore, I decide to dig further into `rkyv`'s implementation of `ArchivedHashMap`. I basically did two refinements:
1. Changed a few generic parameters in `ArchivedHashMap`'s function signature to allow hashmaps with other `RandomState`'s implementation to be archived
2. Created a struct called `VecForHashMap`. User can convert a `Vec<K, V>` to `VecForHashMap<K::Archived, V::Archived, K, V>`. The serialized version of `VecForHashMap` is equivalent to the serialized version of `Hashmap<K, V>`. I created such additional struct because it seems like `ArchivedHashMap` does not use `RandomState` stored in the `HashMap` at all, meaning that under some circumstances where `HashMap` is created purely for serializing and storing, there's no need to create `HashMap` at all, improving performance by not calculating hashes.

I did some tests using following codes:
```rust
fn create_hashmap_traditionally(size: u32) -> Vec<u8> {
    let mut serializer = AlignedSerializer::new(AlignedVec::new());
    let mut hashmap = HashMap::with_capacity(size as usize);
    for i in 0..size {
        hashmap.insert(i, 0u8);
    }
    serializer.serialize_value(&hashmap).unwrap();
    let buffer = serializer.into_inner();
    buffer.into_vec()
}

fn create_hashmap_ahash(size: u32) -> Vec<u8> {
    let mut serializer = AlignedSerializer::new(AlignedVec::new());
    let mut hashmap = HashMap::with_capacity_and_hasher(size as usize, RandomState::new());
    for i in 0..size {
        hashmap.insert(i, 0u8);
    }
    serializer.serialize_value(&hashmap).unwrap();
    let buffer = serializer.into_inner();
    buffer.into_vec()
}

fn create_hashmap_vec(size: u32) -> Vec<u8> {
    let mut serializer = AlignedSerializer::new(AlignedVec::new());
    let vector = (0..size).into_iter().map(|x| (x, 0u8)).collect::<Vec<_>>();
    let hashmap = VecForHashMap::new(vector);
    serializer.serialize_value(&hashmap).unwrap();
    let buffer = serializer.into_inner();
    buffer.into_vec()
}

fn read_hashmap(source: &[u8], size: u32) {
    let archived_value = unsafe {
        archived_root::<HashMap<usize, u8>>(source)
    };
    for i in 0..size {
        assert_eq!(archived_value.get(&i), Some(&0u8))
    }
}
fn criterion_benchmark(c: &mut Criterion) {
    let size = 100;
    c.bench_function("traditional", |b| b.iter(|| read_hashmap(&create_hashmap_traditionally(size), size)));
    c.bench_function("ahash", |b| b.iter(|| read_hashmap(&create_hashmap_ahash(size), size)));
    c.bench_function("vec", |b| b.iter(|| read_hashmap(&create_hashmap_vec(size), size)));
}
```
And the results from my computer is following:
```
traditional             time:   [19.623 us 19.680 us 19.757 us]                         

Found 6 outliers among 100 measurements (6.00%)
  6 (6.00%) high severe

ahash                   time:   [19.343 us 19.421 us 19.537 us]                   

Found 11 outliers among 100 measurements (11.00%)
  3 (3.00%) high mild
  8 (8.00%) high severe

vec                     time:   [16.371 us 16.474 us 16.592 us]                 

Found 21 outliers among 100 measurements (21.00%)
  14 (14.00%) low mild
  5 (5.00%) high mild
  2 (2.00%) high severe
```
As shown, serializing and de-serializing HashMap is ~16% faster when using `VecForHashMap`.

**Possible problems**

I am not a professional rust programmer, and this is actually my first time to create such a "large" pull request, so I might made something wrong in the code or completely misinterpreted logic in this crate, so please tell me if you find any problem in the code. Also, I'm not an English native speaker, so please pay special attention to some documentations I added! Thanks.

In specific, I have following concerns about this PR:
1. Due to some reasons, `ahash` benchmark (above) sometimes perform worse than `traditional` benchmark. I assume that it's caused by `ahash` hashing algorithm instead of something in my code, but I'm not sure since `ahash` claims that it's a "faster drop-in replacement" for HashMap in its documentation.
2. Should I place `VecForHashMap` in `std_impl::chd`? It's technically just a wrapper for `Vec`, but I think it's more related to `ArchivedHashMap` instead of `ArchivedVec`?
3. Should there be a new wrapper called `VecForHashSet`?


moelife-coder:
I reverted the `VecForHashMap`, and now this PR is solely for parameterizing `Hashmap`. I'll probably do a little bit research about the 0.7 branch in this crate and create a PR for the wrapper type in the next few days :)

Issue #126, opened 2021-05-06
Add `alloc` feature for `no_std` containers


djkoloski:
Some `no_std` users can still use global memory allocation, they just can't support the whole standard library. Having a separate `alloc` feature would help round this out so users can opt in to what they can support. It also helps provide the best possible experience for them even though they can't use the whole standard library.

djkoloski:
The bulk of the work is done in 19b93366107a6599d3ce90025f62d0b0b7c2e2c0. Among the things that still need to be fixed are:

- Validation (requires bytecheck fixes)
- With needs to be separated properly
- Tests need to be fixed

These are up next.

djkoloski:
Fixed by cba8af4acfb3e858a27fa67eb590f78870fceae6

PR #149, opened 2021-06-08
Fix build for `specialization` without `std`

PiDelport:
This makes `cargo build -p rkyv --no-default-features --features specialization` work.


djkoloski:
Looks good, thanks for the PR. Linking #126 for visibility, in 0.7 the goal is to split `std` and `alloc` into separate features so we can more easily handle this kind of thing.

Issue #89, opened 2021-03-22
Empty Vec causes SIGSEGV during deserialization


mahou-shoujo:
Hello, the following example produces a SIGSEGV while deserializing the archived data (rkyv 0.4.2, default features).
```
#[test]
fn boom() {
    use rkyv::{
        archived_value,
        de::deserializers::AllocDeserializer,
        ser::{serializers::WriteSerializer, Serializer},
        Archive, Deserialize, Serialize,
    };
    #[derive(Archive, Deserialize, Serialize, Debug, PartialEq)]
    struct Test {
        vec: Vec<i32>,
    }
    let value = Test { vec: vec![] };

    let mut serializer = WriteSerializer::new(Vec::new());
    let pos = serializer
        .serialize_value(&value)
        .expect("failed to serialize value");
    let buf = serializer.into_inner();

    let archived = unsafe { archived_value::<Test>(buf.as_ref(), pos) };

    let mut deserializer = AllocDeserializer;
    let deserialized = archived
        .deserialize(&mut deserializer)
        .expect("failed to deserialize value");
    assert_eq!(deserialized, value);
}
```
A similar test case does work with version 0.3.1 though and it seems like the regression happened in 0.4.0. My rustc profile:
```
debug_assertions
target_arch="x86_64"
target_endian="little"
target_env="gnu"
target_family="unix"
target_feature="fxsr"
target_feature="sse"
target_feature="sse2"
target_os="linux"
target_pointer_width="64"
target_vendor="unknown"
unix
```

djkoloski:
Can you run your buffer through [`bytecheck`](https://github.com/djkoloski/bytecheck) and see if you get an error?

Example:

```rust
// ...

let archived = rkyv::validation::check_archive::<Test>(buf.as_ref(), pos).expect("check bytes failed");

// ...
```

This could be an alignment issue like #83. The docs need a pass to emphasize that buffers need to be properly aligned.

mahou-shoujo:
Yes, `check_archive` also fails with the same memory error (even though it is not an unsafe function). I would think Vec buffers will always have some good alignment on my target but just to get it out of the way I had tried to use `Aligned<[u8; 256]>` as buffer (+ `BufferSerializer`) and still got the same error. 

[...]

PR #153, opened 2021-06-12
Fix crash with jemalloc: don't attempt to allocate zero bytes

devongovett:
When using jemalloc as an allocator, I was seeing crashes similar to the ones reported in #89 when deserializing empty buffers/strings. Looks like another place where this occurs was missed. I based the fix on c803c1b288472284f09611b72914ba5c794206e9 and verified locally that it works.


djkoloski:
Great catch! This PR looks good, sorry for the crash and thanks for the fix! I'll publish this in a new bugfix version shortly.

PR #168, opened 2021-07-09
Fix endofline encoding on unix

remilauzier:


djkoloski:
Looks good, I've updated my `.gitconfig` to `core.autocrlf = true` as well.

PR #190, opened 2021-08-20
Add edition to rustfmt.toml

krl:
If `rustfmt` is run without the `Cargo.toml` context, for example through an emacs mode, it will remove all leading namespace separators `::`. 

A solution to this is to add the epoch in the `rustfmt.toml`


djkoloski:
This seems like a good idea to me. I'll also add the edition to the other `rustfmt.toml`s.

PR #195, opened 2021-09-01
Add optional support for granular endianness

b8591340:
I tried to `#[derive(Archive)]` a `u64_be` newtype and noticed it was missing, so this PR adds support for optional (feature `"endian"`) granular endianness with non-multibyte `impl_primitive` and `impl_atomic`.

I would have asked first, but it was a trivial change. I went by the existing impls assuming they matched `rend`’s API, so try and give a look and see if something needs to be added or adjusted.

rkyv is such a nice project by the way, keep up the great work!

### DIFF
--- a/rkyv/src/impls/core/primitive.rs
+++ b/rkyv/src/impls/core/primitive.rs
@@ -48,9 +48,9 @@ macro_rules! impl_primitive {
             #[cfg(not(any(feature = "archive_le", feature = "archive_be")))]
             type Archived = $type;
             #[cfg(feature = "archive_le")]
-            type Archived = rend::LittleEndian<$type>;
+            type Archived = ::rend::LittleEndian<$type>;
             #[cfg(feature = "archive_be")]
-            type Archived = rend::BigEndian<$type>;
+            type Archived = ::rend::BigEndian<$type>;
 
             impl Archive for $type {
                 type Archived = Archived;
@@ -108,9 +108,9 @@ macro_rules! impl_atomic {
             #[cfg(not(any(feature = "archive_le", feature = "archive_be")))]
             type Archived = $prim;
             #[cfg(feature = "archive_le")]
-            type Archived = rend::LittleEndian<$prim>;
+            type Archived = ::rend::LittleEndian<$prim>;
             #[cfg(feature = "archive_be")]
-            type Archived = rend::BigEndian<$prim>;
+            type Archived = ::rend::BigEndian<$prim>;
 
             type Resolver = ();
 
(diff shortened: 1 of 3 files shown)


djkoloski:
Thanks for the PR! Adding impls for the rend types is a good idea I'll take a look at this as soon as I can. I think that this support may be clearer under a `rend` feature than an `endian` one, but the rest should be fine.

b8591340:
Indeed!

b8591340:
But features and deps cannot have the same name. What do you suggest?

djkoloski:
Ah, that's a good catch. Optional dependencies create a feature by default, so it should be possible to `cfg(feature = "rend")` without explicitly defining a `rend` feature. That way, whenever `rend` gets pulled in, the impls will get pulled in too.

b8591340:
The issue is that currently `rend` is only pulled in when `archive_be` or `archive_le` are enabled. If we want control at the declaration level, which `rend` itself enables, there doesn’t seem another way other than having a new feature; while they are not mutually exclusive, `endian` wouldn’t imply `archive_be` or `_le`, and would still work when opting in to one default order.

djkoloski:
Right, I think there's a little miscommunication. I'm suggesting that we provide these impls whenever `rend` is pulled in, whether that was with the `archive_le`/`archive_be` features or manually with `feature = ["rend"]`. [Here's a bit on how optional dependencies interact with features.](https://doc.rust-lang.org/cargo/reference/features.html#optional-dependencies) This would allow users to get impls for `rend` types manually or automatically when they use archive endianness features.

b8591340:
Oh right, I had totally missed the notion of being able to specify the dependency as a feature externally. My bad. Thanks for taking the time to spell it out, haha; it’s late night here, and I should really try and get some sleep. I’ll move to `feature = "rand"` and fix the rest. Thanks again David!

djkoloski:
Thank you for your help! If it turns out to be more work than you intended to bite off, feel free to let me know and I'll push this PR through to completion. 🙂

b8591340:
The one addition is that I’m re-exporting `rend` when the feature is enabled. I think you might also want to add `rend` to `release_checklist` by itself and with `archive_(be|le)`.

djkoloski:
Awesome, thanks for bearing with me on this one. I'll get this shipped in the next version (0.7.16). I'll also add the new `rend` feature to the docs and tests before then.

b8591340:
Thank you David! It was a pleasure! And thanks again for the great crate. It’s people like you that make the Rust ecosystem a joy to work with.

b8591340:
I was going through the codebase to understand copy optimizations and stumbled upon [`ArchiveCopy` and `ArchiveCopySafe` impls for primitives](https://github.com/rkyv/rkyv/blob/a773ba2c55390ab76b269985dc2009bb1a775253/rkyv/src/copy.rs#L109). You may want to mention in the docs that applying endianness granularly is more efficient than doing so wholesale with `archive_(be|le)`, as one ends up paying the cost only when and when needed.

djkoloski:
I should also provide `ArchiveCopy` and `ArchiveCopySafe` impls when the target endianness matches the archive endianness. I'll take a crack at that.

PR #196, opened 2021-09-01
Fix ArchivedDynMetadata emplace with rend

b8591340:


djkoloski:
Good catch! Sorry about that and thanks for the fix!

djkoloski:
This will be released in `rkyv_dyn` 0.7.6.

Issue #197, opened 2021-09-01
Validation tests failing on Apple M1


b8591340:
```rust
test validation::test_alloc::tests::basic_functionality ... FAILED

failures:

---- validation::test_alloc::tests::basic_functionality stdout ----
thread 'validation::test_alloc::tests::basic_functionality' panicked at 'called `Result::unwrap()`
on an `Err` value: ContextError(ArchiveError(Unaligned { ptr: 0x16b97e316, align: 4 }))’,
rkyv_test/src/validation/test_alloc.rs:128:16
```

I didn’t investigate further but do let me know if I can help.

djkoloski:
I'll get these tested on an M1 machine and see if I can track down the issue.

djkoloski:
Could you run the following snippet? Just want to sanity-check primitive alignment for M1.

```rust
use core::mem::align_of;

fn main() {
    println!("u8: {}", align_of::<u8>());
    println!("u16: {}", align_of::<u16>());
    println!("u32: {}", align_of::<u32>());
    println!("u64: {}", align_of::<u64>());
    println!("u128: {}", align_of::<u128>());
    println!("i8: {}", align_of::<i8>());
    println!("i16: {}", align_of::<i16>());
    println!("i32: {}", align_of::<i32>());
    println!("i64: {}", align_of::<i64>());
    println!("i128: {}", align_of::<i128>());
    println!("f32: {}", align_of::<f32>());
    println!("f64: {}", align_of::<f64>());
    println!("char: {}", align_of::<char>());
    println!("bool: {}", align_of::<bool>());
}
```

The error in question says that an alignment of 4 was expected but it found an alignment of 2. The only way this should be able to happen is if `align_of::<u16>() = 4`, which is unlikely but not impossible.

djkoloski:
Alright, I think this might be a bit of a confusing one. You may have been testing like `cargo test --features "validation size_16"`, but that leaves the default features enabled. One of the default features is `size_32`, which means that both `size_16` and `size_32` were enabled for that test. This should be a compiler error since the features are mutually exclusive, but due to the way the code is structured it ended up just shadowing a few variables. Because having both features enabled is an error, I have added additional compiler errors in 9a5552b132a3acf452db5a645689f6ab2d0d9e5d. I'm not sure that this is the issue that you ran into, but if it is could you try re-running the tests with the latest commit?

b8591340:
```rust
u8: 1
u16: 2
u32: 4
u64: 8
[...]

PR #200, opened 2021-09-02
Fix default features, default workspace members

b8591340:
Enables `--no-default-features` to work across the workspace (#197) by adjusting `default-features` in `rkyv_dyn` and `rkyv_dyn_test` and making `default-members` only crates that need to be.


djkoloski:
Looks good! Thanks for cleaning these up. 🙂

PR #220, opened 2021-11-18
Add ability to use scratch space via a mutable reference.

tobz:
Exactly what it says on the tin.  Related to this: https://github.com/Kixiron/ddshow/blob/master/crates/ddshow-sink/src/writer.rs#L92

Signed-off-by: Toby Lawrence <toby@nuclearfurnace.com>

### DIFF
--- a/rkyv/src/lib.rs
+++ b/rkyv/src/lib.rs
@@ -194,6 +194,10 @@ pub trait Fallible {
     type Error: 'static;
 }
 
+impl<T: ?Sized + Fallible> Fallible for &mut T {
+    type Error = T::Error;
+}
+
 /// A fallible type that cannot produce errors.
 ///
 /// This type can be used to serialize and deserialize types that cannot fail to serialize or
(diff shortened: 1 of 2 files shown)


djkoloski:
I'm not opposed to this change, but you should be able to reuse scratch space by holding onto some memory (in a `Box<[u8]>` for example) and constructing a `BufferScratch<&mut [u8]>` to reuse the space. `HeapScratch` uses [this kind of pattern](https://docs.rs/rkyv/0.7.20/src/rkyv/ser/serializers/alloc.rs.html#113) with a `Box`. Getting the alignment right is a bit tricky, hence the manual allocation, but you should be able to hold onto an `AlignedVec` with the right size.

Sample code:

```rust
struct MyWriter {
    scratch_space: AlignedVec,
}

impl MyWriter {
    fn new() -> Self {
        let mut scratch_space = AlignedVec::with_capacity(4096);
        // not strictly necessary but fast
        unsafe { scratch_space.set_len(scratch_space.capacity()); }
        Self {
            scratch_space,
        }
    }

    fn serialize() {
        // ...
        let mut serializer = CompositeSerializer::new(
            AlignedSerializer::new(&mut self.buffer),
            FallbackScratch::new(BufferScratch::new(&mut self.scratch), AllocScratch::default()),
            Infallible,
        );
        // ...
    }
}
```

Not tested but something like this should do the trick.

tobz:
Hmmmm, that's a fair point.  I guess constructing the fallback/buffer/alloc scratch value should actually be pretty cheap all things considered, so yeah, maybe I'll just roll with that for the meantime. 👍🏻

PR #234, opened 2021-12-17
Add hashbrown feature

Miaxos:
Hello 👋!

There is an implementation for the `std::collections::HashMap` but there is no implementation of `hashbrown::HashMap` which should be the same. I just copied the implementation from [here](https://github.com/Miaxos/rkyv/blob/hashbrown_impl/rkyv/src/impls/std/collections/hash_map.rs) as it's a fixed version of `hashbrown` which is used by the std.

### DIFF
--- a/rkyv/src/impls/hashbrown.rs
+++ b/rkyv/src/impls/hashbrown.rs
@@ -1,0 +1,75 @@
+use crate::{
+    collections::hash_map::{ArchivedHashMap, HashMapResolver},
+    ser::{ScratchSpace, Serializer},
+    Archive, Deserialize, Fallible, Serialize,
+};
+use core::{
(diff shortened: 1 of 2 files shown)


djkoloski:
Thanks for the PR, this looks great. It looks like all the features are already set up for this to work correctly. Once all the checks pass I'll merge this and it will be in the next release (v0.7.27).

hatem-amer:
Any example on how to use `rkyv` with `hashbrown::HashMap` ?

djkoloski:
You should be able to enable support by adding `"hashbrown"` to the `features` array on your rkyv dependency:

```toml
rkyv = { version = "0.7", features = ["hashbrown"] }
```

Does that work for you?

Issue #246, opened 2022-01-29
Implement `PartialEq<Duration>` for `ArchivedDuration`


djkoloski:
This brings `ArchivedDuration` closer in line with other standard library types, which tend to implement comparison traits when possible.

ok-nick:
`Eq` as well

kastolars:
First time contributor here, I'm opening a draft PR but might need some guidance.

djkoloski:
Fixed by 2066b4192dd545dd12cbc2595bd80714150b935e

PR #247, opened 2022-01-29
Implementation for ArchivedDuration PartialEq

kastolars:
Linked issue: https://github.com/rkyv/rkyv/issues/246

Is this appropriate function placement? Where should unit test placement go? Guidance on documentation?


djkoloski:
This is really good! It's a pretty basic impl, but it helps.

> Is this an appropriate function placement?

rkyv has a bit of an odd source layout. All of the archived structures are placed at the top level (e.g. `rkyv::time::ArchivedDuration`), but all of the code that references standard library types goes in `impls`. This makes it easier to make sure that it doesn't get compiled in when `std` is disabled, which would cause a compiler error.

The right file for this is `src/impls/std/time.rs` (which doesn't exist yet).

> Where should unit test placement go?

When possible, unit tests should go at the bottom of the file for the functionality their testing. For simple impls like this, it would probably be okay to forego a unit test. I'd never turn one down though!

> Guidance on documentation?

rkyv uses the `missing_docs` lint, so if the compiler doesn't yell at you then everything is properly documented.

rkyv/src/time.rs
```diff
@@ -109,3 +110,13 @@ impl ArchivedDuration {
         addr_of_mut!((*out).nanos).write(to_archived!(nanos));
     }
 }
+
+impl PartialEq<Duration> for ArchivedDuration {
+    fn eq(&self, other: &Duration) -> bool {
+        return (self.as_nanos() == other.as_nanos()) && (self.as_secs() == other.as_secs())
+    }
+
+    fn ne(&self, other: &Duration) -> bool {
```
djkoloski, line 119: We can use the provided impl for `ne`. It just forwards to the `eq` impl.

rkyv/src/time.rs
```diff
@@ -109,3 +110,13 @@ impl ArchivedDuration {
         addr_of_mut!((*out).nanos).write(to_archived!(nanos));
     }
 }
+
+impl PartialEq<Duration> for ArchivedDuration {
+    fn eq(&self, other: &Duration) -> bool {
```
djkoloski, line 115: Since this is a small function, it's probably a good idea to `#[inline]` it.

rkyv/src/time.rs
```diff
@@ -109,3 +110,13 @@ impl ArchivedDuration {
         addr_of_mut!((*out).nanos).write(to_archived!(nanos));
     }
 }
+
+impl PartialEq<Duration> for ArchivedDuration {
```
djkoloski, line 114: When possible, we also want to implement the reverse relationship. In this case, it's `impl PartialEq<ArchivedDuration> for Duration`. It can just delegate to the other impl.

kastolars:
> > Where should unit test placement go?
> 
> When possible, unit tests should go at the bottom of the file for the functionality their testing. For simple impls like this, it would probably be okay to forego a unit test. I'd never turn one down though!

I'm not sure I can create a unit test without some sort of constructor for `ArchivedDuration` or making its fields public.

djkoloski:
Looks good! This will be ready to merge once CI finishes. If I'm still up I'll merge it before I go to sleep, otherwise I'll do it first thing in the morning.

PR #263, opened 2022-04-12
add impl for with::Map

zyansheep:
Added this to core: https://gist.github.com/djkoloski/880e809f2ba3fcff7d6f2b53a0cb6ef3

Allows stuff like this:
```rust
use rkyv::{Archive, with::{Map, RefAsBox}};
#[derive(Archive)]
struct Example<'a> {
    #[with(Map<RefAsBox>)]
    option: Option<&'a i32>,
    #[with(Map<RefAsBox>)]
    vec: Vec<&'a i32>,
}
```
Its probably not perfect, but its a start.

### DIFF
--- a/rkyv/src/with/core.rs
+++ b/rkyv/src/with/core.rs
@@ -6,21 +6,23 @@ use crate::{
         ArchivedOptionNonZeroU16, ArchivedOptionNonZeroU32, ArchivedOptionNonZeroU64,
         ArchivedOptionNonZeroU8,
     },
-    with::{ArchiveWith, AsBox, DeserializeWith, Inline, Map, Niche, RefAsBox, SerializeWith, Unsafe},
-    Archive, ArchiveUnsized, Deserialize, Fallible, Serialize, SerializeUnsized,
     option::ArchivedOption,
+    ser::{ScratchSpace, Serializer},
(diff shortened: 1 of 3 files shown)


rkyv/src/with/core.rs
```diff
@@ -6,18 +6,183 @@ use crate::{
...
+    hint::unreachable_unchecked,
+    marker::PhantomData,
     num::{
         NonZeroI128, NonZeroI16, NonZeroI32, NonZeroI64, NonZeroI8, NonZeroIsize, NonZeroU128,
         NonZeroU16, NonZeroU32, NonZeroU64, NonZeroU8, NonZeroUsize,
     },
+    ptr,
 };
 
+// Map for Vecs
```
djkoloski, line 29: This part needs to go into `alloc.rs` because `Vec` isn't enabled on `#![no_std]` builds without `extern crate alloc`.

djkoloski (changes requested):
Looks good, just need to shuffle some code around to keep all the build configurations working.

rkyv/src/with/core.rs
zyansheep: fixed

djkoloski (approved):

PR #269, opened 2022-04-28
Update uuid to 1.0

Dylan-DPC:
Updates uuid dependency to the latest released version (1.0) . This is a breaking change if you export the dependency. 

I don't think any more changes are required


seguidor777:
Hello, when are you releasing the next minor version? Since this is a breaking change and I have another dependency (sqlx) which uses uuid < 1.0 I have to lock the rkyv version to `=0.7.37` Due to my case I think it would be better if we had jumped from 0.7.38 to 0.8. Thanks

djkoloski:
I'll add a feature specifically for uuid <1.0 with the next minor release so you can use `uuid_0.8` until sqlx updates.

seguidor777:
Alright, that's a good plan. Thank you

PR #279, opened 2022-06-29
impl Archive, Deserialize, and Serialize for AlignedBytes and AlignedVec

bonsairobo:
The implementation strategy is to derive all of these trait impls. This
was straightforward for AlignedBytes since it is just POD, but it
required a little more effort for AlignedVec. Thankfully, by changing
the representation of AlignedVec to use an internal
Vec<AlignedBytes<16>>, most of the code got simpler. All of the manual
allocations went away, and much of the unsafe code went with it.

This does include a change to some of the test's expectations about what
should be returned by AlignedVec::capacity. Now, capacity is always a
multiple of ALIGNMENT=16. It is unlikely that anyone relied upon the
exact capacity. I think so long as reserve and reserve_exact follow the
same contract of reserving "enough" space for the additional bytes, we
do not need to call this a breaking change.

ArchivedAlignedVec also got some manual trait impls, similar to what
ArchivedVec has, and those impls were mostly copied from those of
ArchivedVec.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -84,16 +84,16 @@ impl AlignedVec {
     /// assert_eq!(vec.capacity(), 16);
     ///
     /// // These are all done without reallocating...
-    /// for i in 0..10 {
+    /// for i in 0..16 {
     ///     vec.push(i);
     /// }
-    /// assert_eq!(vec.len(), 10);
+    /// assert_eq!(vec.len(), 16);
     /// assert_eq!(vec.capacity(), 16);
     ///
     /// // ...but this may make the vector reallocate
-    /// vec.push(11);
-    /// assert_eq!(vec.len(), 11);
-    /// assert!(vec.capacity() >= 11);
+    /// vec.push(16);
+    /// assert_eq!(vec.len(), 17);
+    /// assert!(vec.capacity() >= 17);
     /// ```
     #[inline]
     pub fn with_capacity(capacity: usize) -> Self {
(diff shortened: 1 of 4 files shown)


rkyv/src/util/aligned_vec.rs
```diff
@@ -660,3 +623,139 @@ unsafe impl Send for AlignedVec {}
 unsafe impl Sync for AlignedVec {}
 
 impl Unpin for AlignedVec {}
+
+impl ArchivedAlignedVec {
+    /// Returns the number of elements in the archived vec.
+    #[inline]
+    pub fn len(&self) -> usize {
+        from_archived!(self.len) as usize
```
bonsairobo, line 631: TBH I don't really understand what this is doing. I copied it from `ArchivedVec`.

djkoloski: It's a nasty workaround since sometimes the archived type of `self.len` will be a plain `u32` and other times it will be a `rend` type. It's used correctly here.

rkyv/src/util/aligned_vec.rs
```diff
@@ -2,35 +2,53 @@
...
+/// let arch_foo = unsafe { rkyv::archived_root::<HasAlignedBytes>(&ser_foo) };
+/// assert_eq!(arch_foo.bytes.as_slice(), &[1, 2, 3]);
+/// assert_eq!(arch_foo.bytes.as_ptr().align_offset(16), 0);
+/// ```
+#[derive(Archive, Clone, Deserialize, Serialize)]
+#[archive(crate = "crate")]
 pub struct AlignedVec {
-    ptr: NonNull<u8>,
-    cap: usize,
+    bytes: Vec<AlignedBytes<ALIGNMENT>>,
```
djkoloski, line 45: Because `Vec` contains its own `ptr, cap, len` triple, this will bump `size_of::<AlignedVec>()` up from 24 to 32. It's not necessarily a big problem but it is wasted space since we have redundant information in two `len`s.

bonsairobo:
Closing in favor of #279

PR #280, opened 2022-06-29
impl Archive and Serialize for AlignedBytes and AlignedVec

bonsairobo:
The archived representation of AlignedVec is `ArchivedVec<u8>`. We manually align the serializer to 16 bytes before serializing the slice using the pre-existing `ArchivedVec::resolve_from_slice`.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -1,3 +1,10 @@
+use crate::vec::VecResolver;
+use crate::{
+    ser::{ScratchSpace, Serializer},
+    vec::ArchivedVec,
+    Archive, Archived, Serialize,
+};
+
 #[cfg(not(feature = "std"))]
 use ::alloc::{alloc, boxed::Box, vec::Vec};
 use core::borrow::{Borrow, BorrowMut};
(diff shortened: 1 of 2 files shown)


rkyv/src/util/aligned_vec.rs
```diff
@@ -11,6 +18,25 @@ use core::{
...
+/// struct HasAlignedBytes {
+///     pub bytes: AlignedVec,
+/// }
+///
+/// let mut bytes = AlignedVec::new();
+/// bytes.extend_from_slice(&[1, 2, 3]);
+/// let ser_foo = rkyv::to_bytes::<_, 0>(&HasAlignedBytes { bytes }).unwrap();
+/// let arch_foo = unsafe { rkyv::archived_root::<HasAlignedBytes>(&ser_foo) };
+/// assert_eq!(arch_foo.bytes.as_slice(), &[1, 2, 3]);
+/// assert_eq!(arch_foo.bytes.as_ptr().align_offset(16), 0);
```
bonsairobo, line 38: This actually succeeds even without the call to `Serializer::align`. @djkoloski Do you have an idea for how to force the re-alignment? I tried putting another field on the struct, but it didn't make a difference.

djkoloski: Hmm, we could manually construct a serializer with an intentionally mis-aligned buffer. This would be something like a `BufferSerializer` around `&mut [u8]` pulled out of `AlignedBytes<256>`. Then we would write a single byte so the alignment would be 1 and serializing the `AlignedVec` would require 15 bytes of padding. Asserting before / after should do the trick.

PR #282, opened 2022-07-02
add resize function

### LINKED ISSUE #281: Add `resize` method to `AlignedVec`
Having a `resize` method would make it easier for people using `AlignedVec` as a buffer or write destination.

zyansheep:
Doc tests should cover all the cases.
Closes #281

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -348,6 +348,37 @@ impl AlignedVec {
             }
         }
     }
+    /// Resizes the Vec in-place so that len is equal to new_len.
+    ///
+    /// If new_len is greater than len, the Vec is extended by the difference, with each additional slot filled with value. If new_len is less than len, the Vec is simply truncated.
+    /// 
+    /// # Panics
+    ///
+    /// Panics if the new length exceeds `usize::MAX` bytes.
+    ///
+    /// # Examples
+    /// ```
+    /// use rkyv::AlignedVec;
+    ///
+    /// let mut vec = AlignedVec::new();
+    /// vec.push(3);
+    /// vec.resize(3, 2);
+    /// assert_eq!(vec.as_slice(), &[3, 2, 2]);
+    ///
+    /// let mut vec = AlignedVec::new();
+    /// vec.extend_from_slice(&[1, 2, 3, 4]);
+    /// vec.resize(2, 0);
+    /// assert_eq!(vec.as_slice(), &[1, 2]);
+    /// ```
+    #[inline]
+    pub fn resize(&mut self, new_len: usize, value: u8) {
+        if new_len > self.len {
+            let additional = new_len - self.len;
+            self.reserve(additional);
+            unsafe { std::ptr::write_bytes(self.ptr.as_ptr().add(self.len), value, additional); }
+        }
+        unsafe { self.set_len(new_len); }
+    }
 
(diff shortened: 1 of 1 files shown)


rkyv/src/util/aligned_vec.rs
```diff
@@ -348,6 +348,39 @@ impl AlignedVec {
...
+    /// vec.extend_from_slice(&[1, 2, 3, 4]);
+    /// vec.resize(2, 0);
+    /// assert_eq!(vec.as_slice(), &[1, 2]);
+    /// ```
+    #[inline]
+    pub fn resize(&mut self, new_len: usize, value: u8) {
+        let len = self.len;
+        if new_len > len {
+            self.reserve(new_len - len);
+            let slice = unsafe { slice::from_raw_parts_mut(self.ptr.as_ptr(), new_len) };
```
djkoloski, line 377: I think we should be using `ptr::write_bytes` here since making a slice of the bytes creates a reference to uninitialized memory. `write_bytes` may also result in some more efficient code.

djkoloski (changes requested):

rkyv/src/util/aligned_vec.rs
zyansheep: fixed

rkyv/src/util/aligned_vec.rs
```diff
@@ -348,6 +348,37 @@ impl AlignedVec {
...
+    /// vec.resize(2, 0);
+    /// assert_eq!(vec.as_slice(), &[1, 2]);
+    /// ```
+    #[inline]
+    pub fn resize(&mut self, new_len: usize, value: u8) {
+        let len = self.len;
+        if new_len > len {
+            let additional = new_len - len;
+            self.reserve(additional);
+            unsafe { std::ptr::write_bytes(self.ptr.as_ptr(), value, additional); }
```
djkoloski, line 378: I think this should be `ptr::write_bytes(self.ptr.as_ptr().add(len), value, additional)`.

rkyv/src/util/aligned_vec.rs
```diff
@@ -348,6 +348,38 @@ impl AlignedVec {
...
+    /// assert_eq!(vec.as_slice(), &[3, 2, 2]);
+    ///
+    /// let mut vec = AlignedVec::new();
+    /// vec.extend_from_slice(&[1, 2, 3, 4]);
+    /// vec.resize(2, 0);
+    /// assert_eq!(vec.as_slice(), &[1, 2]);
+    /// ```
+    #[inline]
+    pub fn resize(&mut self, new_len: usize, value: u8) {
+        let len = self.len;
```
djkoloski, line 375: Why save out `len` here? We should be able to just use `self.len` throughout since `reserve` doesn't modify it.

rkyv/src/util/aligned_vec.rs
```diff
@@ -348,6 +348,38 @@ impl AlignedVec {
...
+        if new_len > len {
+            let additional = new_len - len;
+            self.reserve(additional);
+            unsafe { std::ptr::write_bytes(self.ptr.as_ptr().add(len), value, additional); }
+        }
+        self.len = new_len;
```
djkoloski, line 381: Prefer `self.set_len(new_len)` here.

djkoloski (commented):
One last round of minor changes and then I think this will be good to go.

djkoloski (approved):

PR #296, opened 2022-08-21
add BitVec support

truelossless:
This PR adds BitVec support to rkyv.

> bitvec provides a foundational API for bitfields in Rust. It specializes standard-library data structures (slices, arrays, and vectors of bool) to use one-bit-per-bool storage, similar to [std::bitset](https://en.cppreference.com/w/cpp/utility/bitset) and [std::vector](https://en.cppreference.com/w/cpp/container/vector_bool) in C++.

I modified https://github.com/rkyv/rkyv_contrib/pull/3 to directly implement `Archive` on `BitVec` instead of creating a wrapper type. While I was at it, I implemented it for the BitArray type too. Here are a few notes:

- `BitVec<T, O>` now archives to `ArchivedBitVec<Archived<T>, O>` as opposed to https://github.com/rkyv/rkyv_contrib/pull/3 where it would archive to `ArchivedBitVec<T, O>` while using internally an `ArchivedVec<Archived<T>>`, which I think was incorrect behavior.

- `BitArray` doesn't store the bit length because const generics are not stable yet, as explained in the [bitvec docs](https://docs.rs/bitvec/1.0.1/bitvec/array/struct.BitArray.html). This means that an archived `BitArray<usize, O>` will not be equal to a `BitArray<usize, O>` if the `size_XX` cargo feature doesn't match the target's pointer width. This can be a big footgun so maybe removing the `ArchivedBitArray` implementation is the best decision until const generics are available.

- I think implementing `BitBox` would require adding a bitvec feature to the ptr_meta crate. Indeed, we need to store an unsized `BitSlice` in `ArchivedBox`. The [BitSpan docs](https://github.com/bitvecto-rs/bitvec/blob/main/doc/ptr/BitSpan.md), states that the `Pointee` trait could be defined as `(usize, BitIdx<T::Mem>)`. However I don't think it's really important to support it as 99% of the people will be only using `ArchivedBitVec` anyway.

### DIFF
--- a/rkyv/src/impls/bitvec.rs
+++ b/rkyv/src/impls/bitvec.rs
@@ -1,34 +1,15 @@
 use core::convert::{TryFrom, TryInto};
 use core::{marker::PhantomData, ops::Deref};
-
-#[cfg(feature = "bitvec_std")]
+#[cfg(feature = "bitvec_alloc")]
 use crate::vec::{ArchivedVec, VecResolver};
-
 use crate::{
-    out_field,
+    out_field, bitvec::ArchivedBitVec,
     ser::{ScratchSpace, Serializer},
     Archive, Archived, Deserialize, Fallible, Serialize,
 };
 
 use bitvec::{prelude::*, view::BitViewSized};
 
-/// An archived `BitVec`.
-// We also have to store the bit length in the archived `BitVec`.
-// This is because when calling `as_raw_slice` we will get unwanted bits if the `BitVec` bit length is not a multiple of the bit size of T.
-#[cfg(feature = "bitvec_std")]
-#[cfg_attr(feature = "validation", derive(bytecheck::CheckBytes))]
-#[derive(Debug, Eq, Hash, Ord, PartialEq, PartialOrd)]
-pub struct ArchivedBitVec<T = Archived<usize>, O = Lsb0>
-where
-    T: BitStore,
-    O: BitOrder,
-{
-    inner: ArchivedVec<T>,
-    bit_len: Archived<usize>,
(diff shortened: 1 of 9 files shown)


djkoloski:
I moved a few things around to make it easier to cfg code in and out. Running tests currently gives a failure, could you take a look? I ran the tests with `cargo test --no-default-features --features "bitvec size_32 alloc bitvec_alloc" -- bitvec`.

rkyv/src/impls/bitvec.rs
```diff
@@ -173,18 +145,18 @@ mod tests {
...
 
-    #[cfg(feature = "bitvec_std")]
     #[test]
+    #[cfg(feature = "bitvec_alloc")]
     fn bitvec() {
-        use crate::ser::serializers::AllocSerializer;
+        use crate::ser::serializers::CoreSerializer;
 
-        let mut serializer = AllocSerializer::<256>::default();
+        let mut serializer = CoreSerializer::<256, 256>::default();
```
truelossless, line 153: It seems that using `CoreSerializer` causes the test to fail, the archived version just becomes an empty slice. I replaced the `AllocSerializer` by a `CoreSerializer` in `tiny_vec::tests::tiny_vec` and got a similar error.

Did you change the `AllocSerializer` to a `CoreSerializer` by mistake or should it be possible to archive heap-allocated types with it?

djkoloski: Ah, good spot. `CoreSerializer` doesn't work right with `archived_root` out of the box. I'll fix that up, do a once-over, and then this should be good to merge.

PR #304, opened 2022-08-26
Add riscv32im

cpunkzzz:
The project I'm using uses riscv32im instruction set that also doesn't support atomic32/64. Since i vs im vs imc is just differences with extensions added to the standard riscv32, it probably make sense to include them under a single matcher but this works for now


djkoloski:
Thanks for the PR!

PR #306, opened 2022-08-26
Fix typo in Docs::deserialize.md

canewsin:


djkoloski (commented):
This one's not misspelled, "sped up" is the past tense of "speed up".

canewsin:
lack of commas in docs at certain places, makes terms some what confusing.

djkoloski:
I did notice that there appears to be a duplicate "a" in there after "more or less". That sentence could be reworded better though. How about "This provides traditional deserialization, being sped up somewhat by having a very compatible representation."

canewsin:
somewhat and compatible wording is redundant in a single sentence.

canewsin:
https://github.com/rkyv/rkyv/blob/62f0bef9fd3e4c4e762ee574f7dafe05cdfb1320/book_src/zero-copy-deserialization.md?plain=1#L86-L88

These lines also needs rework, especially this 
> it doesn't *scale* with our data

canewsin:
> it doesn't _scale_ with our data

For this 
> This operation is almost zero work, and more importantly data size doesn't increase any work, it's always just a pointer offset and a cast to access our data.

djkoloski:
I think these suggestions could improve the documentation, and I'd be happy to review a PR with them as changes. Since this PR is obsolete, I'm going to close it for now.

PR #308, opened 2022-08-27
Run miri in CI and ignore the failing tests

5225225:
I doubt this will work since the rust.yml workflow won't have miri installed, but that's the command that works.

Pinging @djkoloski can you help me out on where to add the miri here, I've basically never worked with adding it to CI before :)

The `--all-targets` is needed because we fail a doctest because miri doesn't support `ctor`, and we can't put ignore on doctests just under miri, to my knowledge.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -25,5 +25,7 @@ jobs:
       run: cargo build --verbose
     - name: Run tests
       run: cargo test --verbose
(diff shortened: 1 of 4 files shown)


djkoloski:
Thanks for the pointer, I'll switch over to the `rust-toolchain` action and ensure it gets a toolchain that has `miri`.

PR #313, opened 2022-09-09
Add parameterization over the RandomState for IndexMap and IndexSet

truelossless:
This adds the `RandomState` generic type parameter for `IndexMap` and `IndexSet` which were only implemented over the default generic type parameter.

Now the `Serialize`, `Archive` and `Deserialize` traits for `IndexMap` and `IndexSet` should match the behavior of those in `HashMap` and `HashSet`.

The tests are passing and I tried serializing and archiving with a custom hasher, it works fine.


djkoloski:
Thanks for the PR!

PR #315, opened 2022-10-05
Unify documentation of `out_field!` with its uses

### LINKED ISSUE #254: Clarify documentation for  `out_field!`
The names `(fp, fo)` actually stand for `field_pos` and `field_out`, but are referred to in the docs as `offset` and `pointer`. Unifying these naming conventions would help alleviate confusion.

fu5ha:
Fixes #254 by rewording the documentation and example of `out_field!`. Also, changes the naming of the return tuple in the documentation of `Archive` to be more verbose so it's a little less confusing for the uninitiated.

### DIFF
--- a/rkyv/src/macros.rs
+++ b/rkyv/src/macros.rs
@@ -10,14 +10,24 @@ macro_rules! default {
     ($($fn:tt)*) => { $($fn)* };
 }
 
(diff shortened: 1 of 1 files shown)


djkoloski:
Looks like the `mips-unknown-linux-gnu` test is failing. It's not related to this PR so I'll merge it and figure it out in #316.

PR #319, opened 2022-10-12
Add `from_raw_parts` style interfaces to `RelPtr` and `ArchivedBox`

fu5ha:
Discussed the need for this on Discord. Draft for now so I can prototype with it in `protoss` and make sure it works as expected/is necessary.

Also snuck in a ignore `/tmp` directory in `.gitignore`, a moderately common pattern which I use is to have `rust-analyzer` compile into a separate target directory so that it doesn't conflict with regular cargo builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ book
 Cargo.lock
 /target
 /.vscode
+/tmp
--- a/rkyv/src/boxed.rs
+++ b/rkyv/src/boxed.rs
@@ -58,6 +58,25 @@ impl<T: ArchivePointee + ?Sized> ArchivedBox<T> {
         })
     }
 
+    /// Resolves an archived box from a [`BoxResolver`] which contains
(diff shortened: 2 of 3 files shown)


fu5ha:
:ohno: nightly compiler ICE-ing

fu5ha:
Marked ready for review since it seems [to work](https://github.com/fu5ha/protoss/commit/2b71ea0e54765eae939b90d0b8205b9a97c22722) for what I want it to in `protoss`, though hasn't been super thoroughly tested. Maybe worth a MIRI run through :)

Not sure why it would cause the compiler to ICE.... I guess I'll just try rerunning and see if it was a fluke lol.

fu5ha:
Seems the ICE is the same on `rkyv` without this commit on latest ntightly. Filed https://github.com/rust-lang/rust/issues/102978 to track it...

fu5ha:
The relevant ICE was in theory fixed, so should be good to go now :)

Issue #330, opened 2023-01-06
Empty ArchivedHashMap panics


zach-schoenberger:
I found that a get on an empty archived hashmap panics, at least on my system. As best I can tell it has to do with using 0 as the modulus divisor which results in a divide by 0 panic. Below is sample code that reproduces the error.

```
#[derive(serde::Serialize, serde::Deserialize)]
#[serde(rename_all = "snake_case")]
#[derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
#[archive(compare(PartialEq))]
#[archive_attr(derive(bytecheck::CheckBytes, Debug))]
#[derive(Clone, PartialEq, Default)]
pub struct BadMap {
    map: HashMap<u32, u32>,
}

#[test]
fn empty_hashmap_panic() {
    let map = BadMap::default();

    let bytes = rkyv::to_bytes::<_, 256>(&map).unwrap();

    let map_archive = rkyv::check_archived_root::<BadMap>(&bytes[..]).unwrap();

    assert!(map_archive.map.get(&123).is_none());
}
```

and the resulting stack trace:
```
running 1 test
thread 'rkyv::empty_hashmap_panic' panicked at 'attempt to calculate the remainder with a divisor of zero', /Users/zschoenb/.cargo/registry/src/github.com-1ecc6299db9ec823/rkyv-0.7.39/src/collections/hash_index/mod.rs:66:30
stack backtrace:
   0: rust_begin_unwind
             at /rustc/83356b78c4ff3e7d84e977aa6143793545967301/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/83356b78c4ff3e7d84e977aa6143793545967301/library/core/src/panicking.rs:65:14
   2: core::panicking::panic
             at /rustc/83356b78c4ff3e7d84e977aa6143793545967301/library/core/src/panicking.rs:114:5
   3: rkyv::collections::hash_index::ArchivedHashIndex::index
             at /Users/zschoenb/.cargo/registry/src/github.com-1ecc6299db9ec823/rkyv-0.7.39/src/collections/hash_index/mod.rs:66:30
   4: rkyv::collections::hash_map::ArchivedHashMap<K,V>::find
             at /Users/zschoenb/.cargo/registry/src/github.com-1ecc6299db9ec823/rkyv-0.7.39/src/collections/hash_map/mod.rs:62:9
[...]

PR #337, opened 2023-02-08
Don't try to compute remainder of division by zero

s3bk:
Trying to look up an element from an empty HashMap resulted in `% 0`.


djkoloski:
Thanks for the PR! It looks like this fixes #330, I'll close that when this merges.

PR #339, opened 2023-02-09
fix #331

### LINKED ISSUE #331: SIGSEGV from `AlignedVec::reserve_exact`
Very simple repro:

```rust
#[test]
fn repro() {
    use rkyv::AlignedVec;
    let mut bytes = AlignedVec::new();
    bytes.reserve_exact(1);
}
```

Using version `0.7.39`.

s3bk:
`reserve` was already doing it correct.


djkoloski:
Thanks for the PR!

Issue #290, opened 2022-08-04
Re-export all public dependencies


eggyal:
Currently, implementors of (for example) `ArchiveUnsized` must add *the correct version* of `ptr_meta` to their own dependencies.  If `rkyv` were instead to re-export `ptr_meta`, or at very least the relevant parts of it, users could simply use those re-exports instead.

djkoloski:
Are there any guidelines on when dependencies should be re-exported? Doing so may make it easier in some cases, but I'm worried about causing dependency resolution issues.

eggyal:
[Effective Rust, Item 25](https://lurklurk.org/effective-rust/re-export.html) recommends at very least re-exporting any types that appear in your public API.

I personally go a step further and reexport the entire crate in which those types appear, as downstream users may require other parts of the crate to generate or consume those types required by my API.

eggyal:
You may also find [this discussion](https://github.com/rust-lang/api-guidelines/discussions/176) in the official "API guidelines" repo relevant/useful.

PR #340, opened 2023-02-10
reexport `bytecheck::CheckBytes`

mkatychev:
Re-export `CheckBytes` to allow self-contained  `rkyv/validation` feature flag.

This will benefit downstream crates with `rkyv` feature flags,
 example: https://github.com/chronotope/chrono/pull/959

* avoids having the downstream crate from maintaining their own versioning of `bytecheck` 
* simplifies conditional feature flagging

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,7 +54,7 @@
 ```rust
 use rkyv::{Archive, Deserialize, Serialize};
 // bytecheck can be used to validate your data if you want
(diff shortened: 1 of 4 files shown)


mkatychev:
Motivation falls in like with what is mentioned here:
https://github.com/rkyv/rkyv/issues/290

PR #344, opened 2023-02-17
set `bytecheck` crate root when using reexported derive macro

### LINKED ISSUE #343: use of undeclared crate or module `bytecheck` when deriving `ByteCheck`
I introduced an issue in this PR:
https://github.com/rkyv/rkyv/pull/340

the reexport for `rkyv::CheckBytes` is done on the `CheckBytes` trait rather than `bytecheck_derive::CheckBytes`:
https://github.com/rkyv/bytecheck/blob/512eee1cfcfef8eec43a334611cc4bb7e7db1775/bytecheck/src/lib.rs#L169

```rust
use rkyv::{Archive, Deserialize, Serialize};
// bytecheck can be used to validate your data if you want
use rkyv::CheckBytes;

#[derive(Archive, Deserialize, Serialize, Debug, PartialEq)]
// This will generate a PartialEq impl between our unarchived and archived types
#[archive(compare(PartialEq))]
// To use the safe API, you have to derive CheckBytes for the archived type
#[archive_attr(derive(rkyv::CheckBytes, Debug))]
struct Test {
    int: u8,
    string: String,
    option: Option<Vec<i32>>,
}

fn main() {
    let value = Test {
        int: 42,
        string: "hello world".to_string(),
        option: Some(vec![1, 2, 3, 4]),
    };

    // Serializing is as easy as a single function call
    let bytes = rkyv::to_bytes::<_, 256>(&value).unwrap();

    // Or you can customize your serialization for better performance
    // and compatibility with #![no_std] environments
    use rkyv::ser::{serializers::AllocSerializer, Serializer};

    let mut serializer = AllocSerializer::<0>::default();
    serializer.serialize_value(&value).unwrap();
    let bytes = serializer.into_serializer().into_inner();

    // You can use the safe API for fast zero-copy deserialization
    let archived = rkyv::check_archived_root::<Test>(&bytes[..]).unwrap();
    assert_eq!(archived, &value);

    // Or you can use the unsafe API for maximum performance
    let archived = unsafe { rkyv::archived_root::<Test>(&bytes[..]) };
    assert_eq!(archived, &value);

    // And you can always deserialize back to the original type
    let deserialized: Test = archived.deserialize(&mut rkyv::Infallible).unwrap();
    assert_eq!(deserialized, value);
}
```

the code snippet above produces the error below:

``` --> src/main.rs:9:23
  |
9 | #[archive_attr(derive(CheckBytes, Debug))]
  |                       ^^^^^^^^^^ use of undeclared crate or module `bytecheck`
  |
  = note: this error originates in the derive macro `CheckBytes` (in Nightly builds, run with -Z macro-backtrace for more info)
```

samlich:
Fixes #343 depends on https://github.com/rkyv/bytecheck/pull/30

This reexports bytecheck and adds `#[check_bytes(crate = "::rkyv::bytecheck")]` to derived archive types, when `#[archive_attr(derive(CheckBytes))]` is used.

It does not work for `#[archive(as = "Self")]`, then the user must add this manually. Edit: actually, you can just change your import to `use rkyv::{CheckBytes, bytecheck}`. In fact, doing such an import makes this entire commit unnecessary. I still think it's convenient though?

There is some low potential for conflict for existing users, as it will add this attribute even for those not using the reexported macro, as there is no way (that I know of) to determine that exactly. I can't think of a normal situation where this would cause issues, but this is technically a breaking change. (if someone is deriving a type also called CheckBytes from a crate other than bytecheck, this will break their code)

`bytecheck` is removed as a dependency from `rkyv_test`, in order to test this (and it's no longer necessary to depend on it directly).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,7 +22,7 @@ documentation = "https://docs.rs/rkyv"
 repository = "https://github.com/rkyv/rkyv"
 
 [workspace.dependencies]
-bytecheck = "~0.6.8"
+bytecheck = "0.6.10"
 proc-macro2 = "1.0"
 ptr_meta = "~0.1.3"
 quote = "1.0"
--- a/book_src/validation.md
+++ b/book_src/validation.md
@@ -9,10 +9,10 @@ To validate an archive, you first have to derive
 type:
 
 ```rs
-use rkyv::{Archive, CheckBytes, Deserialize, Serialize};
+use rkyv::{Archive, Deserialize, Serialize};
 
 #[derive(Archive, Deserialize, Serialize)]
-#[archive_attr(derive(CheckBytes))]
+#[archive(check_bytes)]
 pub struct Example {
     a: i32,
     b: String,
@@ -20,8 +20,7 @@ pub struct Example {
 }
 ```
 
-The `#[archive_attr(...)]` attribute applies the provided attributes to the archived type. Finally,
-you can use
+The `#[archive(check_bytes)]` attribute derives `CheckBytes` on the archived type. Finally, you can use
 [`check_archived_root`](https://docs.rs/rkyv/0.7.1/rkyv/validation/validators/fn.check_archived_root.html) to
 check an archive and get a reference to the archived value if it was successful:
 
--- a/examples/backwards_compat/Cargo.toml
+++ b/examples/backwards_compat/Cargo.toml
@@ -10,5 +10,4 @@ repository.workspace = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-bytecheck.workspace = true
 rkyv = { path = "../../rkyv", features = ["validation"] }
--- a/examples/backwards_compat/src/main.rs
+++ b/examples/backwards_compat/src/main.rs
@@ -1,10 +1,10 @@
-use bytecheck::CheckBytes;
 use rkyv::{with::AsBox, Archive, Deserialize, Serialize};
 
 // This is the version used by the older client, which can read newer versions
 // from senders.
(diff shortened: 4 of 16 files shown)


djkoloski:
Having a way to default to rkyv's re-exported `bytecheck` is a good idea, but I think parsing the arguments to `archive_attr(derive(...))` is a little too magical. Since `bytecheck` is already integrated so tightly with rkyv, I would prefer a separate attribute. That way, you know that exactly what you pass to `archive_attr` gets applied to the generated type.

Could we add a new argument to the `archive` attribute that does all of this at once? Something like `#[archive(check_bytes)]`, which would add `#[derive(::rkyv::bytecheck::CheckBytes)] #[check_bytes(crate = "::rkyv::bytecheck")]`. I think that would have better ergonomics than what we have right now anyway. Compare:

```rust
// old
#[derive(Archive)]
#[archive_attr(derive(CheckBytes))]
struct Foo { ... }
```

to

```rust
// new
#[derive(Archive)]
#[archive(check_bytes)]
struct Foo { ... }
```

I think that's an ergonomics improvement even in the simplest case, and it would be a huge ergonomics improvement in more complex cases with custom paths and such.

samlich:
That's a much better idea, thanks. I also added a commit to fix compatibility with `bytecheck::Error`'s added Send + Sync bounds, so that the tests would compile.

djkoloski:
I think we'll need to separate the bytecheck version bump into a separate PR. I'd like to do that upgrade when rkyv bumps to 0.8. Since the rest of these changes are semver-compatible with rkyv 0.7, we can ship them sooner.

samlich:
Fixed. I'll make a new PR with the send sync change.

djkoloski:
Thanks for the PR and the patience!

PR #350, opened 2023-02-27
Enforce max capacity for `AlignedVec`

overlookmotel:
`AlignedVec`'s implementation uses the unsafe function `alloc::Layout::from_size_align_unchecked()` in many places. The requirements in terms of alignment are trivially met, but there is an odd [third restriction](https://doc.rust-lang.org/alloc/alloc/struct.Layout.html#method.from_size_align):

> `size`, when rounded up to the nearest multiple of `align`, must not overflow `isize` (i.e., the rounded value must be less than or equal to `isize::MAX`).

As the alignment of `AlignedVec` is 16, this implies the maximum capacity needs to be `isize::MAX - 15`. Currently the code allows capacity of up to `usize::MAX` - roughly twice as large.

This PR:

1. Enforces the correct maximum capacity.
2. Fixes a potential arithmetic overflow in `reserve()`.

In practice, I'm not sure if it's feasible to exceed the maximum anyway, as you'd run out of memory first, at least on 64-bit systems. It's probably possible on 32-bit systems (trying to use over 2 GB of the 4 GB available). But I don't think these changes add any overhead for the hot paths, so there seems little harm in playing it "by the book".

When capacity exceeds `isize::MAX / 2`, rather than increasing by power of 2 (which would exceed max), it caps the capacity at the maximum allowable. This is to allow up to 2 GB capacity on 32-bit systems (rather than 1 GB otherwise).

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -189,14 +189,14 @@ impl AlignedVec {
     /// assert_eq!(vec.capacity(), 10);
     /// vec.shrink_to_fit();
     /// assert!(vec.capacity() >= 3);
+    ///
+    /// vec.clear();
+    /// vec.shrink_to_fit();
+    /// assert!(vec.capacity() == 0);
     /// ```
     #[inline]
     pub fn shrink_to_fit(&mut self) {
-        if self.len == 0 {
-            self.clear()
-        } else {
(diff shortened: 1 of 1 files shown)


overlookmotel:
Thanks for merging this so swiftly.

djkoloski:
Thanks for the PR!

Issue #349, opened 2023-02-27
`AlignedVec::reserve` and `reserve_exact` do the same thing


overlookmotel:
`AlignedVec::reserve_exact()` has the same behavior as `reserve()` - it increases the capacity of the vec to next power of 2.

https://github.com/rkyv/rkyv/blob/2d1c964fda2970d7eb7971805de807cc8996f360/rkyv/src/util/aligned_vec.rs#L522-L529

This is surprising, as it means there's no way to do what the method name would suggest and reserve only a specific number of bytes. It's also out of line with `Vec`'s method of the same name.

Is this intentional? If not, I'd be happy to submit a PR to fix.

djkoloski:
Nope, this is not intentional. I'd be happy for a PR that corrects this. Keep in mind that because the alignment of the allocated memory is high, the allocation size is constrained to a multiple of 16.

overlookmotel:
Thanks for coming back. I'll prepare a PR.

However... I wasn't aware that capacity needs to be a multiple of 16. This constraint is not enforced by `AlignedVec::with_capacity()`, for example.

[alloc::Layout](https://doc.rust-lang.org/alloc/alloc/struct.Layout.html) doesn't seem to demand any such constraint, and I don't think any of the other code in `AlignedVec` relies on this property either.

This was surprising to me. Of course this constraint *does* apply to Rust objects (`(u8, u16)` has size of 4), but as far as I can see from the docs, it's not a requirement for allocations.

I suppose this makes sense because it allows you to request allocations with a very high alignment but much smaller size. e.g. `WebAssembly.Memory` objects created in JS are allocated at addresses which are aligned to `1 << 32` (4 GiB), but their minimum size is 64 KiB. I assume this is so translating a 32-bit WASM pointer to a "real" 64-bit pointer in host can be done with inexpensive bitwise operations.

What do you think?

djkoloski:
My bad, I didn't clarify enough. The `layout` passed to functions like `GlobalAlloc::alloc` _does not_ need to have a size that is a multiple of the alignment. However, the memory returned from those calls will likely be padded out to a multiple of the alignment. So there's nothing wrong with your PR, I just wanted to clarify that in case you were benchmarking or measuring for specific results. Sorry for the confusion!

PR #351, opened 2023-02-27
`AlignedVec::reserve_exact` reserve only what's requested

overlookmotel:
PR for #349.

This assumes that capacity does NOT need to be a multiple of 16 (see [comment](https://github.com/rkyv/rkyv/issues/349#issuecomment-1447209157)).

If that constraint *does* need to be upheld, `reserve()` and `with_capacity()` would also need changes.


djkoloski:
Thanks for the PR!

PR #352, opened 2023-02-28
`AlignedVec` call `handle_alloc_error` on alloc failure

overlookmotel:
[Rust docs for alloc](https://doc.rust-lang.org/alloc/alloc/trait.GlobalAlloc.html#tymethod.alloc) say:

> Clients wishing to abort computation in response to an allocation error are encouraged to call the [`handle_alloc_error`](https://doc.rust-lang.org/alloc/alloc/fn.handle_alloc_error.html) function, rather than directly invoking `panic!` or similar.

`AlignedVec` currently panics if allocation fails. This PR converts panics to `handle_alloc_error` calls.

Sorry for the flood of PRs!

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -132,13 +132,15 @@ impl AlignedVec {
                 "`capacity` cannot exceed isize::MAX - 15"
             );
             let ptr = unsafe {
-                alloc::alloc(alloc::Layout::from_size_align_unchecked(
-                    capacity,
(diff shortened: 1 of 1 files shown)


overlookmotel:
CI fail appears unrelated to content of PR.

djkoloski:
Re-ran the CI job and it passed. Looks like it might be an incremental problem? I hate flaky CI.

djkoloski:
And thanks for the PRs! I appreciate your help!

PR #353, opened 2023-02-28
Custom aligned `AlignedVec` (draft)

overlookmotel:
Exploratory draft only. Please see #354.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -163,7 +163,7 @@ impl<const ALIGNMENT: usize> AlignedByteVec<ALIGNMENT> {
         } else {
             assert!(
                 capacity <= Self::MAX_CAPACITY,
-                "`capacity` cannot exceed isize::MAX - 15"
+                "`capacity` cannot exceed `Self::MAX_CAPACITY`"
             );
             let ptr = unsafe {
                 let layout = alloc::Layout::from_size_align_unchecked(
@@ -421,7 +421,7 @@ impl<const ALIGNMENT: usize> AlignedByteVec<ALIGNMENT> {
     ///
     /// # Panics
     ///
-    /// Panics if the new capacity exceeds `isize::MAX - 15` bytes.
+    /// Panics if the new capacity exceeds `Self::MAX_CAPACITY` bytes.
     ///
     /// # Examples
     /// ```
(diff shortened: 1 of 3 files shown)


rkyv/src/util/aligned_vec.rs
```diff
@@ -48,13 +48,25 @@ use std::{alloc, io};
...
+///
+/// See `AlignedVec`.
+///
+/// ```
+/// use rkyv::{AlignedVec, AlignedVecCustom};
+///
+/// let bytes = AlignedVecCustom::<4096>::with_capacity(1);
+/// assert_eq!(bytes.as_ptr() as usize % 4096, 0);
+/// ```
+pub struct AlignedVecCustom<const ALIGNMENT: usize> {
```
djkoloski, line 63: I'd prefer `AlignedByteVec` with a default `const ALIGNMENT: usize = 16` to `AlignedVecCustom`. We could even `#[deprecate]` `AlignedVec`, which I'd be okay with doing.

overlookmotel: I've made these changes. However, default const parameters aren't very ergonomic without a type alias - you need some type annotation even despite the default:

```rust
// These work
let v = AlignedByteVec::<16>::new();
let v: AlignedByteVec = AlignedByteVec::new();

// Compilation error: type annotations needed for `AlignedByteVec<ALIGNMENT>`
let v = AlignedByteVec::new();
```

[playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=8a161830cd43eafdb8f4860008288a08)

So I think `AlignedVec` needs to stay as a type alias, and personally I don't think there's any need to deprecate it. From my experience, once you go beyond `rkyv::to_bytes` and get into composing a custom serializer, it's quite challenging to figure out all the types. I'm a Rust newbie, so maybe it's just me, but I feel it's better to avoid another compulsory parameter the user has to figure out before their code will compile.

What do you think?

djkoloski: Ah I see the issue with default const parameters. I'm fine with leaving it un-deprecated, I'll probably remove the type alias with the next minor version bump when I can do semver-breaking changes.

djkoloski:
`AllocSerializer`: For now we should leave this type alias alone. Introducing a new const param would be a breaking change, so we should continue to use the type alias here.

`ArchivedVec`: It could be parameterized over the alignment, but I also think we're better off only implementing on the `AlignedVec` (16-aligned) alias for the same reason.

I think these are both good avenues for future work, so we should track them in issues.

rkyv/src/util/aligned_vec.rs
```diff
@@ -863,7 +893,7 @@ impl io::Write for AlignedVec {
 }
 
 // SAFETY: AlignedVec is safe to send to another thread
```
djkoloski, line 895: Nit: update these comments to `AlignedByteVec`. There's also a fair bit of documentation that refers to `AlignedVec` that should now refer to `AlignedByteVec`.

overlookmotel: Yes, this PR is a very rough draft at present. I wasn't at all sure if you'd think it was a good idea.

djkoloski: I think it moves rkyv in a positive direction. 🙂

djkoloski: I think this is the only thing blocking merging right now.

overlookmotel:
Just to say @djkoloski sorry I went silent on this one. My attention moved elsewhere, but still it's rude to just disappear. Glad you finished it off.

djkoloski:
No worries! Such is the nature of open-source. 🙂

PR #355, opened 2023-03-01
Improve performance of `AlignedVec::reserve`

overlookmotel:
While doing some experiments, I was wondering why `AlignedVec` was performing significantly slower than `Vec` when you're just writing slices of `u8`s to it.

I think I figured it out, and this PR bring the performance of `AlignedVec::extend_from_slice` up to parity with `Vec`.

The trick is an optimization in `reserve` [which `Vec` uses](https://doc.rust-lang.org/src/alloc/raw_vec.rs.html#274). When the `Vec`/`AlignedVec` is large compared to to the size of chunks you're writing into it, almost always a call to `reserve` is a no-op - there's already sufficient capacity.

The optimization is to split `reserve` into 2 functions. The fast path for just checking capacity is sufficient and returning is marked `#[inline]`. The more complicated logic for actually extending the vec on the rare occasions that it does need to grow is moved into a separate function marked `#[cold]`. My understanding is that this hints that branch prediction should assume the fast path, and makes `reserve` really small, so it can be easily inlined.

(you probably know how this works better than me, just want to document the explanation for this change as I didn't open an issue for it)

On my test which doesn't involve any serialization, just stuffing loads of data into the `AlignedVec`, this change produces a ~20% speed gain.

```
time:   [12.214 µs 12.224 µs 12.236 µs]
change: [-22.747% -22.464% -22.189%] (p = 0.00 < 0.05)
Performance has improved.
```

The effect is less visible when it's mixed in with other activity. But I still get an approx 1-2% speed increase on the serialization benchmark in RKYV repo.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -383,30 +383,31 @@ impl AlignedVec {
             .len
             .checked_add(additional)
             .expect("cannot reserve a larger AlignedVec");
-        unsafe { self.increase_capacity(new_cap) };
+        unsafe { self.grow_capacity_to(new_cap) };
     }
 
-    /// Increase total capacity of vector to `new_cap` or more.
+    /// Grows total capacity of vector to `new_cap` or more.
     ///
-    /// Actual capacity will be `new_cap` rounded up to next power of 2,
+    /// Capacity after this call will be `new_cap` rounded up to next power of 2,
     /// unless that would exceed maximum capacity, in which case capacity
     /// is capped at the maximum.
     ///
-    /// This is same growth strategy used by `reserve`, and therefore also
-    /// by `push` and `extend_from_slice`.
+    /// This is same growth strategy used by `reserve`, `push` and `extend_from_slice`.
     ///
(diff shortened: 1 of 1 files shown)


overlookmotel:
By the way, this involved refactoring out the growth strategy into a separate method `increase_capacity` which is used by both `reserve` and `push`. I've added this method to the public API, as it'd be useful to me at least.

The method is unsafe, but the contract is trivial for the user to fulfill.

I hope this is OK. If you'd prefer it to stay a private implementation detail, let me know, and I'll do that.

overlookmotel:
I've renamed `increase_capacity` to `grow_capacity_to`, and clarified its purpose in the docs.

I chose `grow_capacity_to` rather than just `grow_capacity` to make it clear the parameter is the total capacity to grow *to*, not the amount to grow *by*. This is in contrast to other methods which are have "by" semantics e.g. `reserve(10)` means "reserve 10 *more*".

djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #357, opened 2023-03-02
Make alloc serializers send again

freqmod:
This summer line 205 (at the top. of the patch) made AllocSerializer !Send which makes it very hard to use with async rust. 

By implementing send explicitly this is avoided. As far as I can see it should not be a problem to implement send the way the ptr field is used in the current code. This is confirmed in the comment in the exchange below too.

https://github.com/rkyv/rkyv/commit/359aa4aec662837b0b6731db194172cf819e4838#r102643633


djkoloski:
Thanks for the PR!

PR #358, opened 2023-03-05
Fix Archive macro for enums whos variants clash with associated types

MaxOhn:
This disambiguates associated types in the `Archive` derive macro for enums so that the macro no longer errors for variants called `Archived` or `Resolver`, e.g.
```rs
#[derive(Archive)]
struct MyEnum {
    Archived,
    Resolver { field: u32 },
}
```


djkoloski:
Thanks for the PR!

PR #359, opened 2023-03-05
Add debug asserts to `AlignedVec` unsafe methods

overlookmotel:
This PR just adds `debug_assert!`s to check the constraints of `AlignedVec`'s unsafe methods are met.


djkoloski:
Thanks for the PR!

PR #360, opened 2023-03-05
Make `AlignedVec::change_capacity` method public

overlookmotel:
`AlignedVec::change_capacity` would be useful to me at least, for 2 reasons:

1. As an optimization when my own code has already determined that capacity needs to grow and by how much.
2. To implement a different growth strategy from exponential growth.

I don't know how you feel about this. If you think public API is large enough already, please feel free to reject this PR.

Builds on PR #359 as touches the same code. Please merge #359 first.

PS I'll get back onto #353 in a bit. Just wanted to get all the more bitty changes in first.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -174,12 +174,22 @@ impl AlignedVec {
 
     /// Change capacity of vector.
     ///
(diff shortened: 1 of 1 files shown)


overlookmotel:
Now rebased on master.

djkoloski:
LGTM, thanks for the PR!

PR #342, opened 2023-02-15
Adds `PartialOrd<&str> for ArchivedString` (and inverse) as well as related trait implementations

Qqwy:
As discussed on Discord, `PartialOrd` between strings and archived strings were missing.
This PR adds:

- `PartialOrd<&str> for ArchivedString`
- `PartialOrd<ArchivedString> for &str`
- `PartialOrd<str> for ArchivedString`
- `PartialOrd<ArchivedString> for str`
- missing `PartialEq<ArchivedString> for str` (only its inverse existed)

PR #364, opened 2023-03-08
Adds PartialOrd for ArchivedString <-> String

Qqwy:
A PartialOrd implementation to compare ArchivedString with String was still missing.
This implementation is important, because without it:

```rust
#[derive(rkyv::Archive, rkyv::Serialize, rkyv::Deserialize)]
#[derive(Debug, PartialEq, Eq, PartialOrd, Ord, Hash)]
#[archive_attr(derive(rkyv::CheckBytes))]
#[archive(compare(PartialEq, PartialOrd))] # <- especially this line
pub enum Value {
    Bool(bool),
    Int(i64),
    String(String)
}
```
will complain about the missing trait implementation.

This PR can be considered a follow up to #342 😊

### DIFF
--- a/rkyv/src/string/mod.rs
+++ b/rkyv/src/string/mod.rs
@@ -215,6 +215,20 @@ impl PartialOrd<ArchivedString> for str {
     }
 }
 
+impl PartialOrd<ArchivedString> for String {
+    #[inline]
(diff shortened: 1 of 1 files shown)


djkoloski:
Thanks for the PR!

PR #370, opened 2023-03-21
Bump bytecheck to 0.7.0

jszwedko:


djkoloski:
This is good but we have to bump rkyv's version as well since bytecheck is part of rkyv's validation API.

jszwedko:
Thanks for looking @djkoloski ! I bumped the version here. I assume it should be minor to match the minor version bump of bytcheck?

Unfortunately there were some breaking changes that affect `rkyv_dyn_test` that ended up being a little beyond me; I tried to follow the compiler errors, but hit a wall in updating `rkyv_dyn_derive`. The relevant `bytecheck` breaking change seems to be:

> https://github.com/rkyv/bytecheck/pull/27 (breaking) Adds Send and Sync bounds to bytecheck's Error type

samlich:
A fix for that is provided in #347, which I accidentally closed by deleting my fork (Verzeihung). If you copy the first commit, it will fix the issue, and then the second one is a little extra bonus which you could consider including, as described in the PR.

jszwedko:
Oh nice, thanks @samlich . I pulled those changes over here. Let me know what you think @djkoloski . I pulled in the extra bit that @samlich mentioned with

> Also added a commit to add Send + Sync for rkyv_dyn's boxed errors, as that seems like a reasonable change in the same spirit, and updating bytecheck required making the (extremely minor) breaking change that CheckDynError no longer implements From<Box<dyn Error>>, only From<Box<dyn Error + Send + Sync>>

But happy to back that out if you prefer.

djkoloski:
Thanks for the PR! I'm going to cut 0.7.41 tonight with the final round of improvements before bumping to 0.8 and merging this PR. I'd like to avoid backporting to 0.7 as much as possible, but it's basically inevitable.

PR #372, opened 2023-03-24
Fix `AlignedVec::shrink_to_fit` when len 0

overlookmotel:
Currently `AlignedVec::shrink_to_fit` can cause UB if `len` is 0.

```rust
let mut vec = AlignedVec::new();
vec.push(1);
vec.pop();
// `len` is now 0
vec.shrink_to_fit();
```

`shrink_to_fit()` results in a call to `alloc::realloc` with `new_size` = 0. [Rust docs](https://doc.rust-lang.org/alloc/alloc/trait.GlobalAlloc.html#method.realloc) says this is UB.

This PR fixes that. Now if `len` is 0, it deallocates instead, returning the vec to state it was in after `AlignedVec::new()`.

### DIFF
--- a/rkyv/src/util/aligned_vec.rs
+++ b/rkyv/src/util/aligned_vec.rs
@@ -193,25 +193,31 @@ impl AlignedVec {
         debug_assert!(new_cap <= Self::MAX_CAPACITY);
         debug_assert!(new_cap >= self.len);
 
-        let new_ptr = if self.cap != 0 {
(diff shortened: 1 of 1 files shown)


djkoloski (approved):

overlookmotel:
It was actually me that introduced this bug in #348 when I fixed another bug. Sorry!

djkoloski:
Thanks for the PR!

PR #378, opened 2023-04-10
Fix regression of version 0.7.40

c410-f3r:
On version 0.7.39 `BufferScratch` was both `Send` + `Sync` but now it is just `Send` making async code harder.

cc #357


c410-f3r:
On a second thought, I am not familiar enough with the internals of `BufferScratch` to affirm that the synchronization of a raw pointer won't generate undefined results.

djkoloski:
I think you're right, reopening.

PR #379, opened 2023-04-13
added `bytecheck?/uuid` to `uuid_std`

mkatychev:
Since `rkyv/validation` does not expose the `CheckBytes` `Uuid`  impl by default, structs containing `Uuid`s are not able to be archived with the new `check_bytes` approach:
```
error[E0277]: the trait bound `uuid::Uuid: rkyv::CheckBytes<DefaultValidator<'_>>` is not satisfied
   --> src/lib.rs:853:29
    |
853 |             .filter_map(|v| rkyv::from_bytes::<MyUuidStruct>(&v).ok())
    |                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `rkyv::CheckBytes<DefaultValidator<'_>>` is not implemented for `uuid::Uuid`
```

### DIFF
--- a/rkyv/Cargo.toml
+++ b/rkyv/Cargo.toml
@@ -52,6 +52,7 @@ validation = ["alloc", "bytecheck", "rend/validation"]
 bitvec_alloc = ["bitvec/alloc"]
 tinyvec_alloc = ["tinyvec/alloc"]
 uuid_std = ["uuid/std"]
+uuid_validation = ["uuid/std", "bytecheck/uuid"]
 
 [package.metadata.docs.rs]
 features = ["validation"]
--- a/rkyv/Cargo.toml
+++ b/rkyv/Cargo.toml
@@ -32,11 +32,11 @@ indexmap = { version = "1.7", optional = true, default-features = false }
 smallvec = { version = "1.7", optional = true, default-features = false }
 arrayvec = { version = "0.7", optional = true, default-features = false }
 tinyvec = { version = "1.5", optional = true, default-features = false }
-uuid = { version = "1.0", optional = true, default-features = false }
+uuid = { version = "1.3", optional = true, default-features = false }
 
 [features]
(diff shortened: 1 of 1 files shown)


rkyv/Cargo.toml
```diff
@@ -52,6 +52,7 @@ validation = ["alloc", "bytecheck", "rend/validation"]
 bitvec_alloc = ["bitvec/alloc"]
 tinyvec_alloc = ["tinyvec/alloc"]
 uuid_std = ["uuid/std"]
+uuid_validation = ["uuid/std", "bytecheck/uuid"]
```
mkatychev, line 55: Not happy introducing a new feature flag, esp. when it goes against the suggestion here:
https://github.com/uuid-rs/uuid/issues/515 but this does get a struct to implement `CheckBytes` when holding a `Uuid`. 

Happy to implement alternatives like rolling `bytecheck/uuid` into `validation`

djkoloski: This should be possible without a new feature flag by adding `"bytecheck?/uuid"` to the `uuid` feature.

rkyv/Cargo.toml
```diff
@@ -45,13 +45,10 @@ copy_unsafe = []
 size_16 = []
 size_32 = []
 size_64 = []
-std = ["alloc", "bytecheck/std", "ptr_meta/std", "rend/std"]
+std = ["alloc", "bytecheck?/std", "ptr_meta/std", "rend?/std", "uuid?/std"]
 strict = ["rkyv_derive/strict"]
+uuid = ["dep:uuid", "bytecheck?/uuid"]
```
mkatychev, line 50: This looks much better!

djkoloski: I had to go check the new cargo resolvers reference. I was pretty sure that there was some way to get this to work in feature resolver v2.

PR #381, opened 2023-04-17
Support `?Sized` serializers for `with::Raw`

djkoloski:
This is just an oversight for the impl, nothing about `Raw` requires that the serializer be `Sized`.


ChillFish8 (commented):
I think this might also apply to CopyOptimize, but maybe that does need to be sized

Issue #384, opened 2023-05-01
Name collision in generics causes  comiple-time error in TypeName derive


edfraenkel:
The following code triggers the problem:
```rust
    #[derive(TypeName)]
    struct Struct<F> {
        a: F,
    }
```
because the generated code uses the variable `F` a second time. From cargo expand we get:
```
error[E0403]: the name `F` is already used for a generic parameter in this item's generic parameters
  --> rkyv_typename\tests\test_derive.rs:86:14
   |
86 |     #[derive(TypeName)]
   |              ^^^^^^^^
   |              |
   |              already used
   |              in this derive macro expansion
87 |     struct Struct<F> {
   |                   - first use of `F`
   |
  ::: ...\src\lib.rs:54:1
   |
54 | pub fn type_name_derive(input: proc_macro::TokenStream) -> proc_macro::TokenStream {
...
```
This can be easily fixed by choosing another identifier (e.g. something like `TYPENAME__F`) which is unlikely to clash with other code. (See also: [procedural macro hygiene](https://doc.rust-lang.org/reference/procedural-macros.html#procedural-macro-hygiene).)

djkoloski:
Fixed by #385

Issue #374, opened 2023-03-31
Lifetimes in generics of TypeName derive macro cause compile errors


edfraenkel:
If there are lifetimes in the generic parameters of a struct definition then using the TypeName derive macro causes compile errors.

The following code fails to compile when the two lines marked with _A_ are enabled:

```rust
#[derive(TypeName)]
struct WithLifetime<
//  'a,              // A compile error
//  B,               // B ok
> {
    life: u32,
//  r#is: &'a u128,  // A compile error
//  hard: B,         // B ok
}
```

In fact there are at least two failure modes:
1) Uncommenting the lines marked with _A_ causes:
```
error: proc-macro derive panicked
 --> src\main.rs:4:10
  |
4 | #[derive(TypeName)]
  |          ^^^^^^^^
  |
  = help: message: called `Option::unwrap()` on a `None` value
```

2) Uncommenting all lines _A_ and _B_ causes:
```
error[E0726]: implicit elided lifetime not allowed here
 --> src\main.rs:4:10
  |
4 | #[derive(TypeName)]
  |          ^^^^^^^^ expected lifetime parameter
  |
  = note: this error originates in the derive macro `TypeName` (in Nightly builds, run with -Z macro-backtrace for more info)
help: indicate the anonymous lifetime
  |
4 | #[derive(T'_, ypeName)]
  |           +++
```

The first failure can be explained by line 98 which causes the panic in line 103 because it incorrectly checks that `results` iterates over at least one element.

https://github.com/rkyv/rkyv/blob/1aea469278da6bbed66c4dc20efd589c0b9059b6/rkyv_typename_derive/src/lib.rs#L98-L103

djkoloski:
`rkyv_typename` has been removed since `rkyv_dyn` has been refactored to no longer require it. Closing this issue.

Issue #383, opened 2023-05-01
Const generics with TypeName derive macro cause compile errors


edfraenkel:
This compile time error has the same root cause as issue #374. It can be triggered by the following code:
```rust
#[derive(TypeName)]
struct Struct<const N: usize> {
    a: u32,
    b: u128,
}
```
which results in:
```
error: proc-macro derive panicked
   --> rkyv_typename\tests\test_derive.rs:170:14
    |
170 |     #[derive(TypeName)]
    |              ^^^^^^^^
    |
    = help: message: called `Option::unwrap()` on a `None` value
```

edfraenkel:
Unlike issue #374 it seems like this issue require writing some additional functionality and would influence some design decisions. There are several possible solutions:

1) _Make the const values part of the default type name:_

Implement it such that a unique representation of the constant occurs in the typename. For the currently supported const-generic types this could be implemented with the `Display` trait. However, despite the fact that this is easy to do, I am not sure that this is the right course of afction because of the following considerations.
In a sense, it is 'accidental' that the `Display` trait works because in general, generating unique const identifiers, is not what it is intended for.
In other words: despite the fact that, at this moment, the Display trait would work nicely, it may break for future versions of rust. Future supported const-generic-types may not use the Display trait at all or may not use the Display trait such that a unique output is generated for every value.

2) _Make a minimal implementation for const generics as part of the type name and extend it as the language is extended:_

Forward compatibility could be mitigated by performing compile time type-checks (probably by using some custom private trait). I am pretty sure this is possible, also with human readable error messages.

3) _Use the `Hash` trait to generate the the automatic derive when const generics are involved_

As far as cosistency is concerned this would seem an option that covers more possible const-generic values. It is likely that many (if not all) new const-generic-types will (allow) the implementation of this trait and it could generate a virtually guaranteed unique typename for each const value. However this would be at the expense of readability of the typename and would imply an extra trait bound when const-generics are used. 

4) _Omit const generics from the type name just like lifetime generics can be omitted_

This would not be a viable solution because const generics, unlike generic lifetimes, can affect the serialization/deserialization process. Also, some const generics are already implemented for built-in types, like fixed length arrays. 

5) _Forbid using the automatic derive for const generics_

In this case all implementations for TypeName would need to be hand-written. Allowing them only in such a limited case might not be derilable because it would require the user to write more boiler plate.

djkoloski:
[...]

PR #385, opened 2023-05-01
Issue 374

edfraenkel:
This fixes issue #374 and adds a possible solution to issue #383 
(both issues share the same root cause) 
Also fixes issue #384


djkoloski:
Looks great, and thank you especially for writing some tests. Thanks for the PR!

Issue #287, opened 2022-07-26
Bug with HashMap and enums


CGamesPlay:
When I run this test, it fails. Some things to note here are that changing "Enum" to "std::option::Option" suppresses the failure, and removing either of the two insert statements suppresses the failure.

```rust
#[cfg(test)]
#[test]
fn rkyv_is_broken() {
    use std::{collections::HashMap, result::Result};

    #[derive(Debug, Default, rkyv::Archive, rkyv::Deserialize, rkyv::Serialize)]
    #[archive_attr(derive(Debug, bytecheck::CheckBytes))]
    pub struct Struct {
        pub cols: HashMap<Enum, ()>,
    }

    #[derive(Debug, PartialEq, Eq, Hash, rkyv::Archive, rkyv::Deserialize, rkyv::Serialize)]
    #[archive_attr(derive(Debug, PartialEq, Eq, Hash, bytecheck::CheckBytes))]
    pub enum Enum {
        None,
        Some(String),
    }

    let mut obj = Struct::default();
    obj.cols.insert(Enum::None, ());
    obj.cols.insert(Enum::Some("".to_owned()), ());
    let bytes = rkyv::to_bytes::<_, 256>(&obj).unwrap();
    let returned = rkyv::check_archived_root::<'_, Struct>(&bytes).unwrap();
}
```

```toml
# Cargo.toml
rkyv = { version = "0.7.39", features = [ "strict", "archive_le", "validation", "uuid" ] }
```

The failure received is:

```
thread 'ext::functions::rkyv_is_broken' panicked at 'called `Result::unwrap()` on an `Err` value: CheckBytesError(StructCheckError { field_name: "cols", inner: InvalidKeyPosition { index: 0 } })', src/ext/functions.rs:125:68
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

djkoloski:
Found the issue with some debugging:
```
Enum discriminant type is isize
ArchivedEnum discriminant type is u8
```
The `Hash` derive generates code that hashes the discriminant of the enum before the contents of the variant (if any). So `Enum` is hashing an `isize` but `ArchivedEnum` is hashing a `u8`. This is a problem because the hashes of the unarchived and archived structs _must_ match for `ArchivedHashMap` to work. This does give us an easy fix though:
```rs
#[repr(u8)]
pub enum Enum {
   ...
}
```
Since that will force the discriminant of `Enum` to also be a `u8`.

[...]

PR #396, opened 2023-06-26
Added get_with method for HashMap

### LINKED ISSUE #392: Custom access APIs for archived maps
`ArchivedHashMap`, `ArchivedBTreeMap`, and other key-bearing data structures struggle with nontrivial keys because of the `K: Borrow<Q>, Q: Hash + Eq + ?Sized` bounds. It's not always possible to implement `Borrow`, so a custom comparison function would be a nice escape hatch. I propose something like:

```rust
fn get_with<Q: Hash + Eq + ?Sized, F: Fn(&K, &Q) -> bool>(&self, key: &Q, compare: F) -> Option<&V>
```

for `ArchivedHashMap`, with a similar API for other structures.

See issue #391 for more context.

GeeWee:
This PR adds a `get_with` method for HashMap to allow for lookup with keys that are impossible to construct based on the normal HashMap rules for `get`, such as tuples of ArchivedStrings.

Fixes #392


rkyv/src/collections/hash_map/mod.rs
```diff
@@ -119,6 +119,24 @@ impl<K, V> ArchivedHashMap<K, V> {
...
+    /// If you are unable to construct a key that fulfils the criteria of `get`, you can use this
+    /// method instead.
+    /// E.g. if you want to use a HashMap with a tuple of String keys, constructing a key for use with `get` might be difficult
+    /// but you can use get_with like this: your_hashmap.get_with(&("my", "key"), |hashmap_key, your_key| &(hashmap_key.0.as_str(), hashmap_key.1.as_str()) == your_key)
+    #[inline]
+    pub fn get_with<Q: Hash + Eq + ?Sized, F: Fn(&K, &Q) -> bool>(&self, key: &Q, comparison_predicate: F) -> Option<&V> {
+        // Code adapted from self.find
+        self.index.index(key).and_then(|i| {
+            let entry = unsafe { self.entry(i) };
+            if comparison_predicate(&entry.key, &key) {
```
GeeWee, line 132: While this works in my extremely simple test case,

I'm a little uncertain about the implementation.
- This obviously requires that the Hash is identical between `&str` and ArchivedString, which we'd need to document. However is this an invariant we're sure holds?

`self.index.index` seems to return an index where the entry MAY be located, that `self.entry` then uncritically finds and uses as the entry. I think this means either:
  - That we're guaranteed that when an ArchivedHashMap is built, that each bucket only contains one entry, which means we can use the bucket and the entry interchangeably?
  - That the current ArchivedHashMap implementation doesn't actually deal properly with hash collisions, as it doesn't probe the entire bucket, but only the first entry.

Or perhaps I'm misunderstanding the flow. Please do advise on whether or not I'm misunderstanding the flow, and whether or not the current implementation of get_with looks fine!

djkoloski: Thanks for the questions:

> # Invariant between hashes
> This implementation obviously requires that the Hash is identical between `&str` and `ArchivedString`

Yes, and I agree that it's a good idea to document this.

> is the Hash between and archived type and an non-archived type an invariant we can guarantee holds?

No, we're not guaranteed this; see #287 for a case where this does not hold. We should expect users to get the hashes to match up when they care about it. It may also be worth documenting that the hashes of derived `Archived*` types may not match by default.

> # Hash collisions
> `self.index.index` seems to return an index where the entry MAY be located

rkyv calculates a perfect hash function for the contained keys during serialization, so no two keys _in the hash map_ will return the same bucket. When a key that is _not_ in the hash map is queried, it may hit a bucket in `find()`. However, `find()` checks the queried and stored keys for equality before returning `Some(index)`, and returns `None` otherwise. With these two methods combined, hash collisions aren't a problem for the hash map.

Aside: the original motivation behind perfect hashing was to avoid empty buckets taking up space when serialized, which it had some mild success in doing. In retrospect though, I think that rkyv _should_ have stuck with SwissTable and dropped or conditioned the SIMD-specific parts of hashbrown. The perfect hash table has more space overhead per bucket that nullifies the space savings for small keys, and calculating the perfect hashing function can be very slow for large hash maps. I think the footgun of slow serialization isn't worth the dubious space savings holistically. I've filed #397 to track this work.

rkyv/src/collections/hash_map/mod.rs
```diff
@@ -119,6 +119,24 @@ impl<K, V> ArchivedHashMap<K, V> {
             .map(|index| unsafe { &self.entry(index).value })
     }
 
+    /// Gets the value associated with the given key, and matching the given predicate F.
+    /// If you are unable to construct a key that fulfils the criteria of `get`, you can use this
+    /// method instead.
+    /// E.g. if you want to use a HashMap with a tuple of String keys, constructing a key for use with `get` might be difficult
+    /// but you can use get_with like this: your_hashmap.get_with(&("my", "key"), |hashmap_key, your_key| &(hashmap_key.0.as_str(), hashmap_key.1.as_str()) == your_key)
```
djkoloski, line 126: Can you move this code into an untested example code block? i.e.

```rust
/// ```no-run
/// let your_hashmap: ArchivedHashMap<(ArchivedString, ArchivedString)> = ...;
/// your_hashmap.get_with(
///     &("my", "key"),
///     |hashmap_key, your_key| &(hashmap_key.0.as_str(), hashmap_key.1.as_str()) == your_key),
/// )
/// ```
```

rkyv/src/collections/hash_map/mod.rs
GeeWee: > Yes, and I agree that it's a good idea to document this.
I should document this at the top-level somewhere right? Seeing as this would also apply to using the `ArchivedHashMap` and hitting #287, even without the new `get_with` method?

> Hash collisions
Okay, makes sense with perfect hashing! Thanks for clarifying - I think that means the implementation I have looks right.

rkyv/src/collections/hash_map/mod.rs
GeeWee: Done!

djkoloski:
Thanks for the patience and the PR!

GeeWee:
Thank you!

PR #402, opened 2023-07-03
create examples manifest

mkatychev:
Mirrored directory structure of https://github.com/tokio-rs/tracing/tree/master/examples and created an `examples` cargo manifest.

This should now allow `cargo-expand` to work appropriately:

```
$ cargo expand --example opcode --manifest-path examples/Cargo.toml
```

This should also allow examples to be checked in CI: `cargo check --examples`


.github/workflows/test.yml
```diff
@@ -83,3 +83,4 @@ jobs:
       - uses: dtolnay/rust-toolchain@stable
       - run: cargo build --verbose
       - run: cargo test --verbose
+      - run: cargo check --examples
```
mkatychev, line 86: this is not actually needed since `cargo test` runs `cargo check` on examples by default.

PR #407, opened 2023-07-29
fix codes checking not exist features: tinyvec_alloc, bitvec_alloc

yjpark:
tinyvec_alloc and bitvec_alloc features were removed when they've included in alloc feature:
    "tinyvec?/alloc", "bitvec?/alloc"

but the codes are still checking with the old features.


djkoloski:
This change ended up getting merged in with a bunch of other work, so closing this as obsolete.

Issue #314, opened 2022-09-30
Add support for tokio's `bytes::Bytes` types.


ChillFish8:
Tokio's `Bytes` structures are probably the second most popular way of manipulating and working with a lot of binary data outside of a `Vec<u8>` itself, it would be really nice to have support for this with Rkyv!

dignifiedquire:
Any thoughts on when this might happen? In an async context this is the easiest way to achieve zero copy for IO, and I would love to combine this with rkyv.

djkoloski:
Adding support wouldn't be a semver-breaking change, so we can add support at any time. If anyone wants to take a crack at it, the impl for `Bytes` should be very similar to the impl for `tinyvec`'s types.

PR #408, opened 2023-08-01
Add tokio bytes support

FallingSnow:
A fairly straight forward addition for using tokio's bytes crate with rkyv. I've tested it with my app as well as a module test I've added, and it seems to be working correctly.

Please let me know if you have any suggestions/changes.

Closes https://github.com/rkyv/rkyv/issues/314


djkoloski:
Thanks for the PR!

PR #411, opened 2023-08-11
Add smol_str support

al8n:
A straightforward addition for using rust-analyzer's [`smol_str`](https://crates.io/crates/smol_str) crate with `rkyv`.

### DIFF
--- a/rkyv/src/impls/smolstr.rs
+++ b/rkyv/src/impls/smolstr.rs
@@ -25,12 +25,7 @@ impl<S: ScratchSpace + Serializer + ?Sized> Serialize<S> for SmolStr {
 impl<D: Fallible + ?Sized> Deserialize<SmolStr, D> for ArchivedString {
     #[inline]
     fn deserialize(&self, _deserializer: &mut D) -> Result<SmolStr, D::Error> {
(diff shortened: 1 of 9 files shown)


rkyv/src/impls/smolstr.rs
```diff
@@ -0,0 +1,64 @@
...
+    #[inline]
+    fn serialize(&self, serializer: &mut S) -> Result<Self::Resolver, S::Error> {
+        ArchivedString::serialize_from_str(self, serializer)
+    }
+}
+
+impl<D: Fallible + ?Sized> Deserialize<SmolStr, D> for ArchivedString {
+    #[inline]
+    fn deserialize(&self, _deserializer: &mut D) -> Result<SmolStr, D::Error> {
+        Ok(if self.as_str().len() <= 23 {
```
djkoloski, line 28: It looks like `SmolStr::new` will do this inline/heap-allocated check for us, so we can just call `SmolStr::new(self.as_str())` here in both cases.

djkoloski (changes requested):

djkoloski (approved):

PR #417, opened 2023-08-29
Add ConstGeneric trait for const generic typename

edfraenkel:
closes issue #383


djkoloski:
`rkyv_typename` has unfortunately been yeeted from tree, in part due to this kind of complexity. I've instead rewritten impls to be registered with IDs instead of hashes of type names, which should be more efficient.

PR #418, opened 2023-08-31
dep(rkyv): don't bring in rend with validation

MaxOhn:
Enabling the `validation` feature currently always brings in the optional `rend` dependency.

All usages of `rend` seem to be either behind the `archive_le` or `archive_be` features so `validation` should not depend on it.


MaxOhn:
This is especially unfortunate since `rend` still depends on `bytecheck ~0.6.7`, "poisoning" all of rkyv's `bytecheck` usages to `<0.7.0`

djkoloski:
Sure, if you need this on a 0.7.* version then target that branch.

MaxOhn:
> Sure, if you need this on a 0.7.* version then target that branch.

No rush on my end, I'm fine with having it fixed for 0.8 😄

djkoloski:
Yep, rend is going to be mandatory now

PR #420, opened 2023-09-05
dep(derive): bumped syn to 2.0

MaxOhn:
Updated `syn` from `1.0` to `2.0`.

Essentially only the attribute parsing was affected. I checked (hopefully) all cases of invalid user input to make sure the errors stay the same or are still appropriate.
The `serde` module was updated by copying from the serde repository again.

There was one `TODO` in the code regarding `archive(repr(...))` which I kept in.


djkoloski:
LGTM, sorry CI broke on this PR

PR #432, opened 2023-09-30
`with_capacity` constructor for certain types

monoid:
Preallocating internal data strucutres make various operations 5-15% faster.


djkoloski:
Thanks for the PR!

Issue #434, opened 2023-10-17
Gate PartialEq and PartialOrd impls from local to foreign types behind `extra_traits`


oeb25:
As discovered in https://github.com/scylladb/scylla-rust-driver/pull/838, having `rkyv` in the compiled dependency tree caused the following code to not compile anymore:
```rs
extern crate rkyv;
fn test() {
  Some("") == Some(&"");
}
```
Producing the following error:
```rs
  --> src/lib.rs:3:14
   |
16 |     Some("") == Some(&"");
   |              ^^ no implementation for `Option<&str> == Option<&&str>`
   |
   = help: the trait `PartialEq<Option<&&str>>` is not implemented for `Option<&str>`
   = help: the following other types implement trait `PartialEq<Rhs>`:
             <Option<T> as PartialEq>
             <Option<U> as PartialEq<ArchivedOption<T>>>
```

By minimizing `rkyv` I managed to reduce it to this: ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=d82e564231efa8bdb1c3d359df55b9ef))
```rs
struct ArchivedOption<T>(T);

impl<T: PartialEq<U>, U> PartialEq<ArchivedOption<T>> for Option<U> {
    fn eq(&self, _: &ArchivedOption<T>) -> bool {
        todo!()
    }
}

fn test() {
    Some("") == Some(&"");
}
```

Which corresponds to the following part of the rkyv. Removing this `impl PartialEq` alone removes the error.
https://github.com/rkyv/rkyv/blob/d592789c7a21570455682c095486916a6e7b61a8/rkyv/src/option.rs#L212-L217

I have created [this repo](https://github.com/oeb25/rkyv-partial-eq-error-reproduction) which reproduces the error.

djkoloski:
That’s surprising. I’m predisposed to saying that this is a Rust issue, but I can investigate some more before sending this upstream.

oeb25:
Yes I agree that this seems like some sort of trait resolution issue.

I'll leave a minimal reproduction here as well: ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9a3cfc2279467aa2959f9daacf8983cb))

```rs
struct Wrap<T>(T);

impl<T: PartialEq> PartialEq<Wrap<T>> for Option<T> {
    fn eq(&self, _: &Wrap<T>) -> bool {
        true
    }
}

fn test() {
    Some(()) == Some(&());
}
```

[...]

PR #435, opened 2023-10-17
Remove problematic `impl PartialEq` for `Option<U>` with `ArchivedOption`

oeb25:
This PR fixes the issues of #434 by removing the offending `impl PartialEq` for `ArchivedOption`.

I believe this is a breaking change as shown by https://github.com/rkyv/rkyv/commit/ef8b5ac5d9dd4ed74a354a7ce0a46dbfde14ab02, so I understand if you don't want to merge this.

---

I also attempted to remove one of the generics of the impl, but it still causes the error to occur: ([playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9ccc7e56e76ddce02c37caceac060277))

```rs
impl<T: PartialEq> PartialEq<ArchivedOption<T>> for Option<T> {
    fn eq(&self, other: &ArchivedOption<T>) -> bool {
        other.eq(self)
    }
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,5 +31,5 @@ quote = "1.0"
 rend = { version = "0.5", git = "https://github.com/rkyv/rend", branch = "master", default-features = false }
 rkyv_derive = { version = "=0.8.0", path = "rkyv_derive" }
 seahash = "4.0"
-syn = "1.0"
+syn = "2.0"
 wasm-bindgen-test = "0.3"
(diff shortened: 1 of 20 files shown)


djkoloski:
I put this behind the new `extra_traits` feature in 8c1a8865337689a817da321b04e45b672560d797 and listed you as a co-author of the commit. Closing this as pseudo-merged.

PR #439, opened 2023-11-02
Add ThinVec and triomphe::Arc support

novacrazy:
Mostly just copied from the `sync::Arc` and `TinyVec` implementations.

Unfortunately `Deserialize` can only be implemented for `triomphe::Arc` with `T: Sized`, but oh well. I'm unsure if any of the other `triomphe` types can be used with rkyv, they're kind of restrictive.


novacrazy:
Would it be possible to backport these two and the `smol_str` implementation back to the 0.7 release?

djkoloski:
I rebased and updated this PR, but wasn't able to push to the PR branch. Instead, I created a co-authored commit at fdea22d099c9d2d38b4aa4913edceb283294efe7 that merges this functionality. Thanks for the PR!

PR #442, opened 2023-11-17
0.7 fix atomics esp

ProfFan:
Add the ESP32 and S3 to the list without 64-bit atomics


djkoloski:
You may also need to PR against some of rkyv's dependencies to get your platforms working with 0.7. In the upcoming 0.8 version of rkyv, atomics support has been switched to use the new `target_has_atomic` config.

PR #444, opened 2023-12-01
Fix pointers being displayed incorrectly in `ArchiveError`

oisyn:
There seems to be an issue in the `Display` implementation for `ArchiveError`. Specifically, all pointers are printed using `{:p}`, but because all matches are references, this displays the address of the pointer itself, rather than the _value_ of the pointer.

This PR fixes this by explicitely deferencing the references.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,5 +31,5 @@ quote = "1.0"
 rend = { version = "0.5", git = "https://github.com/rkyv/rend", branch = "master", default-features = false }
 rkyv_derive = { version = "=0.8.0", path = "rkyv_derive" }
 seahash = "4.0"
-syn = "1.0"
+syn = "2.0"
 wasm-bindgen-test = "0.3"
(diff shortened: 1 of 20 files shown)


MarijnS95:
Maybe `{:?}` should be used, which prints the value of a pointer regardless how many times it is (implicitly) borrowed:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=649c1d5fd0bea02e118a24820a4c242f

oisyn:
I considered that, but that's relying on the debug formatter. How guaranteed is it that the debug formatting of a pointer is just a bare hexadecimal value like `{:p}`?

djkoloski:
Fixed up this PR for merge, but wasn't able to push to the upstream branch. I created a co-authored commit at 8591d48c51c082d214c16d0cb6f25bdc0f59bad9 instead, so this is "merged".

PR #453, opened 2023-12-26
Implement Clone for ArchivedString.

gz:
Couldn't derive Clone on ArchivedString, mostly everything had a Clone implementation already except for the `ArchivedStringRepr`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,5 +31,5 @@ quote = "1.0"
 rend = { version = "0.5", git = "https://github.com/rkyv/rend", branch = "master", default-features = false }
 rkyv_derive = { version = "=0.8.0", path = "rkyv_derive" }
 seahash = "4.0"
-syn = "1.0"
+syn = "2.0"
 wasm-bindgen-test = "0.3"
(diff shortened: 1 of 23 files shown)


djkoloski:
This structure shouldn't be safe to `Clone` since it (sneakily) contains a relative pointer. Was there a reason why you wanted to `.clone()` it rather than deserialize it into a `std::String`?

gz:
Yes, I see what you mean about the RelPtr I missed that in the PR.

We're trying to implement an LSM tree data-structure to store indexes on-disk, we have some situations where it would be nice to Clone already archived data: e.g., the most common one is if we're merging two files on-disk files to form a new file. Another application would involve using the archived format more throughout the code-base where operators working on individual data entries want to e.g., take a row and put it in another index data-structure (like a vector etc.)

I guess one way to make ArchivedString clonable would be to extend the enum with a 3rd variant that's big enough/heap allocated? I'm not sure if that would work with/be aligned with the rest of rkyv (fwiw the same limitation appears to be with ArchivedVec which would need a similar strategy to implement Clone)? if you would consider something like that I can work on it as a PR, otherwise we'll find a way to work around and remove the Clone trait constraint on the Archived types.

djkoloski:
I think a good option here would be a deserialize-on-write type, kind of like how `Cow` behaves. You might have to bound `T: Deserialize<some_concrete_deserializer>` but you could also parameterize it over `D: Default`.

gz:
Makes sense, closing this for now.

PR #455, opened 2024-01-04
PartialOrd fix for ArchivedVec.

gz:
Without it, I couldn't get e.g., this to compile since it couldn't compare `Vec<String>` with `ArchivedVec<ArchivedString>`:

```
#[derive(Debug, Clone, SizeOf, Archive, Serialize, Deserialize)]
#[archive_attr(derive(Clone, Ord, Eq, PartialEq, PartialOrd))]
#[archive(compare(PartialEq, PartialOrd))]
pub struct PersonalNetworkGkgEntry {
    pub people: Vec<String>,
}
```

(FYI I hope that this partial_ord impl is actually correct since I haven't written any PartialOrd for slices/vectors before so definitely good to double check :O)


gz:
Closing this with a new PR that is based on main.

PR #461, opened 2024-01-09
Default to `pointer_width_32` if pointer width is unspecified

### LINKED ISSUE #460: size_[16,32,64] features and library development
I am developing a data structures library, and I want to have optional support for rkyv.
So, I added rkyv to my Cargo.toml:
```toml
[dependencies.rkyv]
version = "0.7"
optional = true
default-features = false
```
My library is `no_std`, so I've disabled the default features.

However, `cargo build --all-features` no longer works, failing with:
> error: one of ["size_16", "size_32", or "size_64"] features must be enabled

Because these features are all mutually exclusive (exactly one must be enabled), my library should not be enabling any of them, to allow downstream consumers to choose the repr that they want.
However, that means that I can't build my code unless I pick one of these features.

I've tried also adding rkyv to my `[dev-dependencies]`, which does make `cargo build --all-features --all-targets` work (because `--all-targets` builds unit tests which pulls in dev deps).
However, I still can't do anything else (like build documentation etc) without adding `--features rkyv/size_32` to all of my commands, which is not ideal.
The best I've been able to do is add `__rkyv_hack = ["rkyv/size_32"]` to my crate `[features]`, which works, but is quite hacky.

Is there a better way to do this? Did I miss something in the documentation?

---

While this use case could probably benefit from [Pre-RFC: Mutually-excusive, global features](https://internals.rust-lang.org/t/pre-rfc-mutually-excusive-global-features/19618), such a concept doesn't currently exist :sob:

Sky9x:
WIP

Fixes #460

### DIFF
--- a/rkyv/src/lib.rs
+++ b/rkyv/src/lib.rs
@@ -208,31 +208,43 @@ core::compiler_error!(
 
 // Check pointer width feature flag settings
 
-#[cfg(not(any(
+#[cfg(all(
     feature = "pointer_width_16",
     feature = "pointer_width_32",
-    feature = "pointer_width_64",
-)))]
-core::compile_error!(
-    "\"pointer_width_16\", \"pointer_width_32\", or \"pointer_width_64\" must \
-    be enabled features"
-);
-
-#[cfg(all(feature = "pointer_width_16", feature = "pointer_width_32",))]
+    not(feature = "pointer_width_64")
+))]
 core::compile_error!(
     "\"pointer_width_16\" and \"pointer_width_32\" are mutually-exclusive \
     features. You may need to set `default-features = false` or compile with \
     `--no-default-features`."
 );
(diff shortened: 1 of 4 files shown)


Sky9x:
maybe a better approach would be to check if none of them are set in a build script then manually set feature=pointer_width_32

Sky9x:
or what if we remove the pointer_width_32 feature entirely (just have 16 & 64 for overriding the default)

djkoloski:
We should still have a way for end users to explicitly declare "this must be 32-bit" so that they don't accidentally get overridden.

Sky9x:
> We should still have a way for end users to explicitly declare "this must be 32-bit" so that they don't accidentally get overridden.

yes yes you're right

Sky9x:
that works, thanks

Alxandr:
Any chance this can be back-ported to stable (and released)?

PR #466, opened 2024-01-25
bug: fix validation errors with pointer values

BurntSushi:
I've been hitting an issue where validation fails (and I don't know why), and I've been getting this error:

    error: Reading from cache archive failed: context error: unaligned
    pointer: ptr 0x7fff559663b8 unaligned for alignment 4

Which is quite confusing because `0x7fff559663b8 % 4 == 0`. What was happening instead was that the address of the pointer was being printed instead of the pointer itself.

There are a few ways to fix this, but I think this is a case where match egonomics is harming things. That is, we match on `*self` and specifically insert borrows where necessary. That way, everything else is guaranteed to be used by value. Now my error message (which I'm still trying to track down) is this:

    error: Reading from cache archive failed: context error: unaligned
    pointer: ptr 0x7f3f4556d585 unaligned for alignment 4

I wish it could be more useful (like mentioning the trace to the data type it is tripping on, although that seems hard to do without sacrificing something important), but at least the message itself is correct now.

### DIFF
--- a/rkyv/src/validation/validators/archive.rs
+++ b/rkyv/src/validation/validators/archive.rs
@@ -101,7 +101,7 @@ unsafe impl Sync for ArchiveError {}
 
 impl fmt::Display for ArchiveError {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        match self {
+        match *self {
             ArchiveError::Overflow { base, offset } => write!(
                 f,
                 "relative pointer overflowed: base {:p} offset {}",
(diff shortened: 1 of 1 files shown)


rkyv/src/validation/validators/archive.rs
```diff
@@ -296,6 +305,11 @@ impl<'a> ArchiveContext for ArchiveValidator<'a> {
                 actual_align: self.alignment(),
             })
         } else if (data_address as usize) & (layout.align() - 1) != 0 {
+            dbg!((
+                data_address as usize,
+                layout.align(),
+                (data_address as usize) & (layout.align() - 1)
+            ));
```
djkoloski, line 312: Rogue `dbg!`?

djkoloski:
Feel free to get in touch through the discord as well if you want any help tracking down validation errors, I'd be glad to help.

rkyv/src/validation/validators/archive.rs
BurntSushi: Whoops, nice catch!

BurntSushi: Fixed

BurntSushi:
Ah I didn't know about the Discord! I'll check it out. Thank you.

djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #479, opened 2024-03-05
Fix compiler warnings

matthiasgoergens:
I went through all the warnings that `cargo check` produces on my machine, and fixed them.

### DIFF
--- a/rkyv/src/collections/swiss_table/index_map.rs
+++ b/rkyv/src/collections/swiss_table/index_map.rs
@@ -21,7 +21,6 @@ use crate::{
         HashTableResolver,
     },
     hash::{hash_value, FxHasher64},
-    out_field,
     primitive::ArchivedUsize,
     ser::{Allocator, Writer, WriterExt as _},
     Portable, RelPtr, Serialize,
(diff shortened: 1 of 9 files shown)


matthiasgoergens:
When I run `cargo check --all-targets` I get a few more warnings, but I'm not quite sure what to do about them, so I'm ignoring them for now.  An example:

```console
warning: non-local `impl` definition, they should be avoided as they go against expectation
   --> rkyv_test/src/lib.rs:135:27
    |
135 |         #[derive(Archive, Deserialize, Serialize, Debug, PartialEq)]
    |                           ^^^^^^^^^^^
    |
    = help: move this `impl` block outside the of the current function `archive_raw_identifiers`
    = note: an `impl` definition is non-local if it is nested inside an item and neither the type nor the trait are at the same nesting level as the `impl` block
    = note: one exception to the rule are anon-const (`const _: () = { ... }`) at top-level module and anon-const at the same nesting as the trait or type
    = note: this lint may become deny-by-default in the edition 2024 and higher, see the tracking issue <https://github.com/rust-lang/rust/issues/120363>
    = note: the derive macro `Deserialize` may come from an old version of the `rkyv_derive` crate, try updating your dependency with `cargo update -p rkyv_derive`
    = note: this warning originates in the derive macro `Deserialize` (in Nightly builds, run with -Z macro-backtrace for more info)
```

djkoloski:
This is an open issue with the nightly toolchain. Let's wait for this issue to settle before we start addressing the warnings: https://github.com/rust-lang/rust/issues/121708

djkoloski:
Waited long enough, sorry for the delay!

matthiasgoergens:
No worries. Thanks for merging.

PR #483, opened 2024-03-24
Option<Float> Niche optimization

### LINKED ISSUE #283: Add a non-NaN/non-Inf niched float wrapper
Being able to take advantage of float niches when you know the floats are non-NaN and/or non-Inf would be really useful

novacrazy:
Implements #283 as both a raw `Option<f32> -> ArchivedOptionF32` low-level wrapper and as a `with::Niche`/`ArchiveWith` implementation using `noisy_float::NoisyFloat` for safety. 

`noisy_float` is an optional dependency, and a simple `FloatChecker` trait has been included with `option_float` for non-`noisy_float` usage, but this is automatically implemented for all `noisy_float::FloatChecker` types if enabled. 

~~Some of the generated docs are bit off with the lowercase type names, so I may fix those later.~~ Fixed.

Future work may be to include all `F: Float` types instead of only `f32` and `f64`, but that would require work with rend or others to implement `Float` or whatnot. 

Starting this as a draft so I can review it again in the morning and fix anything I missed.

Closes #283

### DIFF
--- a/rkyv/Cargo.toml
+++ b/rkyv/Cargo.toml
@@ -37,6 +37,7 @@ uuid = { version = "1.3", optional = true, default-features = false }
 bytes = { version = "1.4.0", optional = true, default-features = false }
 thin-vec = { version = "0.2.12", optional = true, default-features = false }
 triomphe = { version = "0.1", optional = true, default-features = false }
+noisy_float = { version = "0.2.0", optional = true, default-features = false }
 
 [features]
 default = ["little_endian", "pointer_width_32", "std", "bytecheck"]
--- a/rkyv/src/impls/mod.rs
+++ b/rkyv/src/impls/mod.rs
@@ -24,6 +24,8 @@ mod bytes;
 mod hashbrown;
 #[cfg(feature = "indexmap")]
 mod indexmap;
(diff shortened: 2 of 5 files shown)


djkoloski:
Hmm, I see the problem here; we don't really have a niche-able float type to bridge from. I really like your approach of adding a checker type parameter that decides whether a value is valid or not, I think that we might even want to generalize it. I could see something like this:

```rust
union NichedOption<T, D> {
    some: T,
    none: (),
}

trait Decider<T> {
    /// # Safety
    ///
    /// `ptr` must be non-null, properly-aligned, and be valid for reads. `ptr` does not have to point to a valid `T`.
    unsafe fn is_some(ptr: *const T) -> bool;
}
```

Which should allow us to generalize the `Niche` wrapper:

```rust
struct Niche<D>(PhantomData<D>);

struct Zero;

impl Decider<NonZeroU32> for Zero { .. }

struct NaN;

impl Decider<f32> for NaN { .. }

// etc.
```

It would be really nice to unify all of this niching code. What do you think? I understand if you'd rather not pivot like this, since it's a pretty significant change in scope.

djkoloski:
Closing this since #554 has been merged

PR #484, opened 2024-03-24
Implement CheckBytes on ArchivedOptionBox and ArchivedOptionNonZeroN, plus `ArchivedOptionNonZeroN::take()`

### LINKED ISSUE #368: add CheckBytes impls for archived niche types
`ArchivedOptionNonZero*` and `ArchivedOptionBox` are currently lacking a `CheckBytes` implementation.

### LINKED ISSUE #474: Add API to set niched `Option<NonZero*>` types to `None`
Current [ArchivedOptionNonZeroU32](https://docs.rs/rkyv/latest/rkyv/niche/option_nonzero/struct.ArchivedOptionNonZeroU32.html#) supports limited mutation using its multiple `mut` or `insert` methods. But there is no way for us to reset it None.

Similar problem exists with `ArchivedOption` but it being an enum, we can easily set it to `ArchivedOption::None`. But no such option is available in case of `ArchivedOptionNonZeroU32`.

Please provide something like a `take` method and set with `Option` method so that we can set the value to `None` if needed.

novacrazy:
As the title says.

Closes #474
Closes #368

The `ArchivedOptionNonZeroN::take()` is a bit extra, but I was in the file.

### DIFF
--- a/rkyv/src/niche/option_box.rs
+++ b/rkyv/src/niche/option_box.rs
@@ -40,11 +40,10 @@ const _: () = {
             value: *const Self,
             context: &mut C,
         ) -> Result<(), C::Error> {
-            // bypass `ArchivedBox::check_bytes` in favor of `RelPtr::check_bytes`
-            CheckBytes::check_bytes(
-                core::ptr::addr_of!((*value).inner.ptr),
-                context,
-            )?;
+            // bypass `ArchivedBox::check_bytes` in favor of `RelPtr::check_bytes`.
+            // both ArchivedOptionBox and ArchivedBox are transparent to the RelPtr,
+            // so casting to RelPtr is safe.
+            RelPtr::check_bytes(value.cast(), context)?;
 
             // verify with null check
             Verify::verify(unsafe { &*value }, context)
--- a/rkyv/src/niche/option_box.rs
+++ b/rkyv/src/niche/option_box.rs
@@ -15,15 +15,59 @@ use crate::{
 /// It uses less space by storing the `None` variant as a null pointer.
 #[derive(Portable)]
 #[archive(crate)]
-#[cfg_attr(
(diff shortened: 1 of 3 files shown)


novacrazy:
I need to test the ArchivedOptionBox more.

novacrazy:
Is it okay to leave that test where it is or should I move it to `rkyv_test`? 

I did have to reimplement it to bypass `ArchivedBox::verify`, as that fails with null pointers required for the `None` variant. It's a bit awkward, but should work fine.

rkyv/src/niche/option_box.rs
```diff
@@ -20,6 +20,54 @@ pub struct ArchivedOptionBox<T: ArchivePointee + ?Sized> {
...
+    where
+        T: ArchivePointee + CheckBytes<C> + LayoutRaw + ?Sized,
+        T::ArchivedMetadata: CheckBytes<C>,
+        C: Fallible + ArchiveContext + ?Sized,
+        C::Error: Error,
+    {
+        #[inline]
+        fn verify(&self, context: &mut C) -> Result<(), C::Error> {
+            if self.inner.is_null() {
+                return Ok(());
```
djkoloski, line 63: Nit: I would prefer `if .. { Ok(()) } else { self.inner.verify(context) }` to avoid the explicit `return`.

djkoloski (approved):

djkoloski:
> Is it okay to leave that test where it is or should I move it to `rkyv_test`?

In the future I'd like to move the tests out of rkyv_test and back into the main codebase, so this is actually a step in the right direction.

novacrazy:
Should be good to go. You're welcome to squash this.

djkoloski:
Thanks for the PR! Sorry for the delay!

PR #500, opened 2024-05-05
Added range and reversability to ArchiveBTreeMap iterators

john-sharratt:
This is really needed to do anything useful with the BTreeMap rather than just an expensive sorter, having the ability to use ranges in Log(O(n)) time allows for interesting use-cases such as very fast loading of BTreeMaps which can then be used by rendering engines immediately with memory copies or reading the entire tree. Also highly useful for very optimized databases that would need to do range scans on file backed data.

- Regression tested locally
- Added unit tests
- Maximum backwards compatibility
- Performance tweaks for fast 
- Retained hint size on forward iterator
- Removed ExactSizeIterator (this won't work with the ranges)

P.s. This was much harder to write than I thought it would be....oh well!

### DIFF
--- a/rkyv/src/collections/btree_map/mod.rs
+++ b/rkyv/src/collections/btree_map/mod.rs
@@ -934,7 +934,7 @@ impl<'a, K, V> Iterator for RawIter<'a, K, V> {
         match &self.run {
             RawIterRun::RunningExact { remaining, .. } => (*remaining, Some(*remaining)),
             RawIterRun::Running { .. } | RawIterRun::RunningInit { .. } => (0, None),
-            RawIterRun::LastOne { .. } => (1, Some(1)),
+            RawIterRun::LastOne { .. } => (0, Some(1)),
             RawIterRun::Finished => (0, Some(0)),
         }
     }
--- a/rkyv/src/collections/btree_map/mod.rs
+++ b/rkyv/src/collections/btree_map/mod.rs
@@ -418,7 +418,7 @@ impl<K, V> ArchivedBTreeMap<K, V> {
     }
 
     /// Constructs a double-ended iterator over a sub-range of elements in the map.
-    /// #[inline]
+    #[inline]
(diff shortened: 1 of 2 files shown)


rkyv/src/collections/btree_map/mod.rs
```diff
@@ -365,6 +474,77 @@ impl<K, V> ArchivedBTreeMap<K, V> {
...
+                    Ok(i) => unsafe { &*inner.tail[i].ptr.as_ptr() },
+                    Err(0) => unsafe { &*inner.header.ptr.as_ptr() },
+                    Err(i) => unsafe { &*inner.tail[i - 1].ptr.as_ptr() }
+                };
+
+                let mut next_classified: ClassifiedNode<'a, K, V> = next.classify();
+                if let ClassifiedNode::Leaf(leaf) = next_classified {
+                    if std::ptr::addr_eq(leaf, cur.0) {
+                        next = match scan {
+                            Ok(0) => unsafe { &*inner.header.ptr.as_ptr() },
```
djkoloski, line 527: It looks like this case will fail to find the previous entry if the current entry is the first in a subtree. Here's an example:
```
    root
    /  \
left   right
/  \   /  \
a  b  c   d
```
If the current entry is the first one of `b`, then the last inner node before it will be `left`. However, there is no leaf node before `b` since it's also the first leaf node of `left`. The right entry to return is the last entry of `c`.

I could be misunderstanding since the tests are passing, but it's also possible that the test input size needs to be larger for this failure mode to express.

I think the only way to do this without allocating would be to write this recursively: each node finds the exact entry for `current` and then returns the previous entry or "beginning of subtree". If a non-root node returns "beginning of subtree" then the parent gets the last entry of the next branch. If the root returns "beginning of subtree" then there is no previous entry.

djkoloski (changes requested):

djkoloski:
I've ceased feature development for rkyv 0.7, so I'm closing this PR.

PR #502, opened 2024-05-14
Fix partial ord

bojanserafimov:
Hi devs :wave: 

I noticed a correctness bug in `#[archive(compare(PartialOrd))]`. I added a failing test case, and fixed it (I think). I'd be happy to address any comments you have on the code and get a fix merged if you're taking contributions

### DIFF
--- a/rkyv_derive/src/archive/struct.rs
+++ b/rkyv_derive/src/archive/struct.rs
@@ -422,7 +422,7 @@ fn generate_partial_ord_impl(
                 #(
                     match other.#members.partial_cmp(&self.#members) {
                         Some(::core::cmp::Ordering::Equal) => (),
-                        x => return x.map(|o| o.reverse()),
+                        x => return x.map(::core::cmp::Ordering::reverse),
                     }
                 )*
                 Some(::core::cmp::Ordering::Equal)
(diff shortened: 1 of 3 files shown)


rkyv_derive/src/archive/struct.rs
```diff
@@ -422,7 +422,7 @@ fn generate_partial_ord_impl(
                 #(
                     match other.#members.partial_cmp(&self.#members) {
                         Some(::core::cmp::Ordering::Equal) => (),
-                        x => return x,
+                        x => return x.map(|o| o.reverse()),
```
djkoloski, line 425:
```suggestion
                        x => return x.map(::core::cmp::Ordering::reverse),
```
I just prefer to avoid writing closures when I can.

rkyv_derive/src/archive/struct.rs
```diff
@@ -436,7 +436,7 @@ fn generate_partial_ord_impl(
                 &self,
                 other: &#name #ty_generics,
             ) -> Option<::core::cmp::Ordering> {
-                other.partial_cmp(self)
+                other.partial_cmp(self).map(|o| o.reverse())
```
djkoloski, line 439:
```suggestion
                other.partial_cmp(self).map(::core::cmp::Ordering::reverse)
```

djkoloski (changes requested):
Nice catch, and thanks for the fix. I suspect that this may also be a problem with enums, could you also add a test case for enums and see if it passes? Enums also support generating `PartialOrd` impls.

djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #448, opened 2023-12-16
Implement PartialOrd between Option/ArchivedOption.

gz:
This implements PartialOrd for comparing Option<T> and ArchivedOption<T>.

PR #525, opened 2024-08-24
Add PartialEq/PartialOrd impls for foreign vec types

novacrazy:
Ran into a case of:
```rust
#[derive(rkyv::Archive)]
#[archive(compare(PartialEq))]
struct Inner {
    a: i32,
}

#[derive(rkyv::Archive)]
#[archive(compare(PartialEq))]
struct Outer {
    a: ThinVec<Inner>,
}
```

This PR fixes the missing `PartialEq` impls. I saw that `Vec` also implements `PartialOrd` with `ArchivedVec`, so I could also add those impls if desired, but the implementation is more complex than these so I skipped it.

### DIFF
--- a/rkyv/src/impls/arrayvec.rs
+++ b/rkyv/src/impls/arrayvec.rs
@@ -51,6 +51,24 @@ where
     }
 }
 
+impl<T, U, const CAP: usize> PartialEq<ArchivedVec<U>> for ArrayVec<T, CAP>
+where
+    T: PartialEq<U>,
+{
+    fn eq(&self, other: &ArchivedVec<U>) -> bool {
+        self.as_slice().eq(other.as_slice())
+    }
+}
+
+impl<T, U, const CAP: usize> PartialEq<ArrayVec<U, CAP>> for ArchivedVec<T>
+where
+    T: PartialEq<U>,
+{
+    fn eq(&self, other: &ArrayVec<U, CAP>) -> bool {
+        self.as_slice().eq(other.as_slice())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use arrayvec::ArrayVec;
--- a/rkyv/src/impls/arrayvec.rs
+++ b/rkyv/src/impls/arrayvec.rs
@@ -51,21 +51,27 @@ where
     }
 }
 
-impl<T, U, const CAP: usize> PartialEq<ArchivedVec<U>> for ArrayVec<T, CAP>
-where
-    T: PartialEq<U>,
-{
-    fn eq(&self, other: &ArchivedVec<U>) -> bool {
-        self.as_slice().eq(other.as_slice())
-    }
-}
-
 impl<T, U, const CAP: usize> PartialEq<ArrayVec<U, CAP>> for ArchivedVec<T>
 where
     T: PartialEq<U>,
 {
     fn eq(&self, other: &ArrayVec<U, CAP>) -> bool {
         self.as_slice().eq(other.as_slice())
+    }
+}
+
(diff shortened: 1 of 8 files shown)


rkyv/src/impls/arrayvec.rs
```diff
@@ -51,6 +51,24 @@ where
     }
 }
 
+impl<T, U, const CAP: usize> PartialEq<ArchivedVec<U>> for ArrayVec<T, CAP>
```
djkoloski, line 54: I'm hesitant to provide generic `PartialEq` impls for generic external types after #434. See [this playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=aa694f15084ca15d1542bea1041e69ae) for an example. I'm fine with adding more `PartialEq` impls which are `for ArchivedType`s though.

gz: I'm curious about this actually wasn't there an `extra_trait` feature? https://github.com/rkyv/rkyv/pull/448 it's here but I can't find it anymore in 0.7 nor 0.8?

djkoloski: There was, but I removed it because I wanted to have a more consistent API surface (not enable/disable the reverse impls with a feature). To compensate somewhat, there are new `get_with` methods implemented for all of the maps. I'd like to support your use case as best as I can, but also would prefer to avoid an `extra_traits` feature if possible.

novacrazy:
Removed those problematic `PartialEq` impls and added a unified lexicographical `PartialOrd` impl on many vec-likes.

rkyv/src/impls/thin_vec.rs
```diff
@@ -58,9 +79,23 @@ mod tests {
...
-        );
+        roundtrip_with(&ThinVec::<i32>::from_iter([10, 20, 40, 80]), |a, b| {
+            assert_eq!(**a, **b)
+        });
+    }
+
+    #[test]
+    fn test_partial_eq() {
+        #[derive(crate::Archive)]
+        #[archive(crate = crate, compare(PartialEq))]
```
djkoloski, line 90:
```suggestion
        #[rkyv(crate, compare(PartialEq))]
```

rkyv/src/impls/thin_vec.rs
```diff
@@ -58,9 +79,23 @@ mod tests {
...
-            |a, b| assert_eq!(**a, **b),
-        );
+        roundtrip_with(&ThinVec::<i32>::from_iter([10, 20, 40, 80]), |a, b| {
+            assert_eq!(**a, **b)
+        });
+    }
+
+    #[test]
+    fn test_partial_eq() {
+        #[derive(crate::Archive)]
```
djkoloski, line 89:
```suggestion
        #[derive(Archive)]
```

djkoloski: And add `use crate::Archive;`

rkyv/src/impls/thin_vec.rs
```diff
@@ -58,9 +79,23 @@ mod tests {
...
+        #[archive(crate = crate, compare(PartialEq))]
+        struct Inner {
+            a: i32,
+        }
+
+        #[derive(crate::Archive)]
+        #[archive(crate = crate, compare(PartialEq))]
```
djkoloski, line 96:
```suggestion
        #[derive(Archive)]
        #[rkyv(crate, compare(PartialEq))]
```

djkoloski (changes requested):

djkoloski:
LGTM after changes

novacrazy:
Should be good now.

djkoloski (approved):

PR #529, opened 2024-08-28
Fix Archive derive for enums

novacrazy:
Seems to have been missing the `#rkyv_path` and assumed `Archive` trait was in scope. Now matches the semantics of struct derives.


djkoloski (approved):

djkoloski:
Thank you! 😊

PR #530, opened 2024-08-29
Implement Positional/Writer/Sharing for references

novacrazy:
My RPC system uses a lot of async streams of objects that are individual/isolated but I could reuse buffers between items after the binary data is written out. Just passing in `&mut my_writer`/`&mut my_share` to a Serializer would be nice, compared to having to take apart the serializer every iteration with `into_raw_parts` to gain access to the buffers again, and having to use Options with `take` and whatnot for stuff in the loop. 

This also adds a `.clear()` method to `Share` to reset it but keep the allocated memory.

### DIFF
--- a/rkyv/src/ser/writer/mod.rs
+++ b/rkyv/src/ser/writer/mod.rs
@@ -164,10 +164,9 @@ pub trait WriterExt<E>: Writer<E> {
 
 impl<T, E> WriterExt<E> for T where T: Writer<E> + ?Sized {}
 
-#[cfg(test)]
+#[cfg(all(test, feature = "alloc"))]
 mod tests {
-    use crate::api::high::to_bytes_in;
-    use crate::util::AlignedVec;
+    use crate::{api::high::to_bytes_in, util::AlignedVec};
     use rend::{u16_le, u32_le};
 
     #[test]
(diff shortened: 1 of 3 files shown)


rkyv/src/ser/writer/mod.rs
```diff
@@ -136,3 +163,31 @@ pub trait WriterExt<E>: Writer<E> {
 }
 
 impl<T, E> WriterExt<E> for T where T: Writer<E> + ?Sized {}
+
+#[cfg(all(test, feature = "alloc"))]
```
djkoloski, line 167:
```suggestion
#[cfg(test)]
```

rkyv/src/ser/writer/mod.rs
```diff
@@ -136,3 +163,31 @@ pub trait WriterExt<E>: Writer<E> {
...
+mod tests {
+    use rend::{u16_le, u32_le};
+
+    use crate::{api::high::to_bytes_in, util::AlignedVec};
+
+    #[test]
+    fn reusable_writer() {
```
djkoloski, line 174:
```suggestion
    #[cfg(feature = "alloc")]
    #[test]
    fn reusable_writer() {
        use rend::{u16_le, u32_le};

        use crate::{api::high::to_bytes_in, util::AlignedVec};

```

djkoloski (changes requested):
LGTM, just one minor nit

novacrazy:
I learned Github has a feature to apply suggestions right here on the web interface. Nice.

djkoloski (approved):

PR #532, opened 2024-09-01
postgres-types feature implementation

novacrazy:
As discussed on Discord. The `postgres-types` crate doesn't keep up that well with dependencies or adding new ones, sometimes, so it's probably just easier to add support here. The implementations are very simple at least.

Should I also open a PR on `rend` to add support for primitives? Personally I don't use archived primitives often in my SQL.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           - bytecheck
         external:
           - ''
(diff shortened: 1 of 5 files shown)


djkoloski (changes requested):
Could you also add `postgres-types` to the list of crate features in `lib.rs`? Then LGTM; thank you for also updating the CI build!

> Should I also open a PR on rend to add support for primitives? Personally I don't use archived primitives often in my SQL.

I'd say PR when you need it

djkoloski (approved):

PR #535, opened 2024-09-02
Forward std feature to rancor

MaxOhn:
Same as #534 but for the `std` feature so that `StdError` gets implemented for `rancor` types.


djkoloski:
Thank you!

PR #536, opened 2024-09-02
Update macro docs

MaxOhn:
Updated some docs for the `Archive`, `Serialize`, and `Deserialize` derive macros.

### DIFF
--- a/rkyv_derive/src/attributes.rs
+++ b/rkyv_derive/src/attributes.rs
@@ -193,24 +193,24 @@ impl Attributes {
             if let Some(ref ident) = result.archived {
                 return Err(Error::new_spanned(
                     ident,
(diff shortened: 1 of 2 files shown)


MaxOhn:
https://github.com/rkyv/rkyv/blob/9d1b30f31580ca07d5df0c43892e89b079ac8e7d/rkyv_derive/src/lib.rs#L64

Maybe this should now read "enable safe accessing"?

djkoloski (approved):

djkoloski:
> https://github.com/rkyv/rkyv/blob/9d1b30f31580ca07d5df0c43892e89b079ac8e7d/rkyv_derive/src/lib.rs#L64
> 
> Maybe this should now read "enable safe accessing"?

Both would probably be the most explicit: "enable safe accessing and deserialization".

PR #539, opened 2024-09-05
Implement `*With` traits via derive macros

MaxOhn:
This adds the ability to put `#[rkyv(remote = Type)]` on container definitions `T` which implements
- `ArchiveWith<Type> for T` with the `Archive` macro
- `SerializeWith<Type, _> for T` with the `Serialize` macro
- `DeserializeWith<Archived<T>, Type, _> for T` with the `Deserialize` macro

The `#[with(...)]` attribute on fields is extended to accept a comma-separated list of either `Type` (just like before) or `remote(...)` which configures the `*With` implementations.

The `remote(...)` item accepts a comma-separated list of
- `Type` which needs to be specified if the remote field's type is different
- `with = Type` to specify a with-wrapper to convert from the remote field's type
- `getter = Path` to specify a path to a function of signature `fn(&RemoteType) -> &RemoteFieldType` or `fn(&RemoteType) -> RemoteFieldType` which is required if the remote field is private.

The namings are inspired by serde's way of [handling remote types](https://serde.rs/remote-derive.html).

Deriving `Deserialize` for `T` with `#[rkyv(remote = R)]` requires a `From<T> for R` implementation.

## Checklist

- [x] Add tests (`/rkyv/tests/derive.rs`)
- [x] Parse macro attributes
- [x] Implement traits
- [ ] Add documentation
- [ ] Add examples

### DIFF
--- a/rkyv/tests/derive.rs
+++ b/rkyv/tests/derive.rs
@@ -2,16 +2,19 @@ use std::{fmt::Debug, marker::PhantomData, mem::MaybeUninit};
 
 use rancor::{Fallible, Panic, ResultExt, Source, Strategy};
 use rkyv::{
-    api::low::LowSerializer, ser::{allocator::SubAllocator, writer::Buffer, Writer}, with::{ArchiveWith, DeserializeWith, Map, SerializeWith}, Archive, Archived, Deserialize, Place, Resolver, Serialize
+    api::low::LowSerializer,
+    ser::{allocator::SubAllocator, writer::Buffer, Writer},
+    with::{ArchiveWith, DeserializeWith, Map, SerializeWith},
+    Archive, Archived, Deserialize, Place, Resolver, Serialize,
 };
+
+type ArchivedWith<F, T> = <F as ArchiveWith<T>>::Archived;
 
 fn roundtrip<F, T>(remote: &T)
 where
     F: ArchiveWith<T, Archived: CheckedArchived>
         + for<'a, 'b> SerializeWith<T, Serializer<'a, 'b>>
-        + DeserializeWith< <F as ArchiveWith<T>>::Archived, T, Strategy<(),
-          Panic>,
-        >,
+        + DeserializeWith<ArchivedWith<F, T>, T, Strategy<(), Panic>>,
     T: Debug + PartialEq,
 {
     let mut bytes = [0_u8; 256];
@@ -152,12 +155,8 @@ fn unnamed_struct() {
         }
     }
 
-    let remote = Remote(
-        42,
-        PhantomData,
-        Some(Foo::default()),
-        Some(Foo::default()),
-    );
+    let remote =
+        Remote(42, PhantomData, Some(Foo::default()), Some(Foo::default()));
 
     roundtrip::<Example<i32>, _>(&remote);
     roundtrip_partial::<Partial<i32>, _>(&remote);
@@ -217,7 +216,7 @@ fn full_enum() {
                 Example::C { a, b, c } => Remote::C {
                     a,
                     b: b.map(From::from),
-                    c: c.map(From::from)
+                    c: c.map(From::from),
                 },
             }
         }
@@ -250,7 +249,7 @@ fn full_enum() {
                     a: *a,
                     b: b.as_ref().copied().map(From::from),
                     c: Some(Foo::default()),
-                }
+                },
             }
         }
     }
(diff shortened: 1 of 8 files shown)


MaxOhn:
Am happy to comply with an requests for change and also no rush by any means 😃 

Will add some documentation and examples next. Perhaps I could write a page in the book too.

MaxOhn:
While checking on implementations for partial types I was wondering how enums should be handled.

I've settled on
- All variants of the remote enum must have a variant in the new type
- Any variant field may be omitted
- All variants keep the same type e.g. a tuple variant in the remote enum must stay a tuple variant, even if all fields are removed

```rs
enum Remote {
    A,
    B(i32),
    C { a: bool, b: [u8; 2] }
}

#[derive(Archive, Serialize, Deserialize)]
#[rkyv(remote = Remote)]
enum Example {
    A,
    B(),
    C { b: [u8; 2] }
}

// Fill in missing fields manually for the `Deserialize` impl
impl From<Example> for Remote {
    fn from(value: Example) -> Self {
        match value {
            Example::A => Remote::A,
            Example::B() => Remote::B(1234),
            Example::C { b } => Remote::C { a: false, b }
        }
    }
}
```

Worth noting that serde doesn't seem to be supporting partial types at all with it's remote implementation.

MaxOhn:
A current issue is that the compiler complains if the native `with` wrapper is the same as the remote one.

```rs
struct Remote { byte: Option<NonZeroU8> }

type OtherNiche;

#[derive(Archive)]
#[rkyv(remote = remote::Remote)]
struct RemoteRef {
    #[with(Niche, remote(with = OtherNiche))]
    pub byte: Option<NonZeroU8>,
}
```

This results in the errors
```
overflow evaluating the requirement `<Niche as ArchiveWith<Option<NonZero<u8>>>>::Archived == _`
overflow evaluating the requirement `<Niche as ArchiveWith<Option<NonZero<u8>>>>::Resolver == _`
```
due to the added trait bound
```
OtherNiche: ::rkyv::with::ArchiveWith<
    Option<NonZeroU8>,
    Archived = <Niche as ::rkyv::with::ArchiveWith<Option<NonZeroU8>>>::Archived,
    Resolver = <Niche as ::rkyv::with::ArchiveWith<Option<NonZeroU8>>>::Resolver,
>
```

djkoloski:
Do you have a use case in mind for keeping `with` and `with(remote(with))` separate? I think it may be clearer for us to not generate `Archive`/`Serialize`/`Deserialize` impls when the type is marked `remote`.

MaxOhn:
>I think it may be clearer for us to not generate `Archive`/`Serialize`/`Deserialize` impls when the type is marked `remote`.

That actually sounds fine to me.

The main downside I can think of is that the 0.8 utility functions don't work well with `ArchiveWith` types anymore so when going doing `Archived -deserialize> Remote -serialize> Bytes` they're basically required to use a self-defined wrapper to use the derived implementations.

It would make macro usage a little more simpler though since it would just be `#[with(Wrapper, getter = path::to::fn)]` and make our macro implementations a little easier too.

> Do you have a use case in mind for keeping `with` and `with(remote(with))` separate?

Had roughly a scenario like this in mind:

```rs
struct Remote {
    field: u32,
}

struct MaybeNonZero;

impl ArchiveWith<u32> for MaybeNonZero {
    type Archived = <Niche as ArchiveWith<NonZeroU32>>;
    ...
}

#[derive(Archive)]
#[rkyv(remote = Remote)]
struct Local {
    #[with(Niche, remote(u32, with = MaybeNonZero))]
    field: Option<NonZeroU32>,
}
```

But yea I wouldn't mind dropping the `Archive`/`Serialize`/`Deserialize` impls 👍

djkoloski:
> The main downside I can think of is that the 0.8 utility functions don't work well with `ArchiveWith` types anymore

I agree, and I actually think the whole crate would benefit from `_with` variants of those utility functions. Would adding those help solve this pain point? It would be so nice to just add back the `With` type but I really don't want to regress on the bad error message.

MaxOhn:
>I actually think the whole crate would benefit from `_with` variants of those utility functions.

Initially I expected the `deserialize_with` and `serialize_with` functions to do just that 😄
Their names are justified though so it's tough to also add functions for `with`-utility. Perhaps an `rkyv::with::api` module or rather `rkyv::api::with`? Or maybe the wrapper workaround is sufficient / it's not actually that much of a pain point so it's fine to maybe put utility functions into `rkyv_util`? Am unsure myself.

MaxOhn:
Continued in #544 since a rewrite was simpler than rebasing

PR #540, opened 2024-09-06
Fix `Result::Ok` variant path

Handy-caT:
In `rkyv_derive` full path is defined for `Result` but not for `Result::Ok`. It leads to possible name conflicts like with `eyre` crate. If user uses `use eyre::*;`,  `eyre::Ok` is also imported. In this case derive macro errors.


djkoloski:
Thanks for the PR!

PR #543, opened 2024-09-07
Future-proof external crate compatibility

novacrazy:
When adding `smol_str` 0.3 support I realized later that after `rkyv` 0.8 it may be a while until the next version is released, so breaking changes to the features would be avoided, and thus dependencies could go very stale. This is an attempt to resolve that early by versioning the feature flags directly, similar to how crates like `postgres-types` do. 

If you don't mind, I also moved external crate support to their own directory since it just looks cleaner IMHO.


djkoloski (approved):

djkoloski:
Thanks! This is a great change.

PR #544, opened 2024-09-08
Implement `*With` traits via derive macros (continued)

MaxOhn:
Continuation of #539 

Adding `#[rkyv(remote = Type)]` on container definitions will implement `ArchiveWith`/`SerializeWith`/`DeserializeWith` instead of `Archive`/`Serialize`/`Deserialize`.

The attribute `#[rkyv(getter = path::to::function)]` can be used to access private fields of the remote type.

### DIFF
--- a/rkyv/tests/derive.rs
+++ b/rkyv/tests/derive.rs
@@ -131,6 +131,12 @@ fn unit_struct() {
     #[rkyv(remote = Remote)]
     struct Example;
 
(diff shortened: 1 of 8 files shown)


djkoloski (approved):

djkoloski:
Thanks so much! This was a big feature, and I'm happy it will be in the next major release.

PR #545, opened 2024-09-09
Add example for remote types

### LINKED ISSUE #446: Write example for using remote derive
Hello, I'm trying to use `rkvy` in a `struct` which has custom type coming from external library, which provides custom serialize/deserialize, but not sure how to wire them. Here is example Code.

```rust 
#[derive(Clone, Serialize, Deserialize, PartialEq, Archive)]
pub struct Store {
   pub id: i32,

   pub looker: SomeCustomLooker // coming from external library <- how to use custom functions here??
}
```

As it happens, external library is providing this functions: 

```rust
serialize(val: SomeCustomLooker) -> Vec<u8>
deserialize(val: Vec<u8>) -> Result<SomeCustomLooker, ....>
```

In `serde` and `bincode`, I can do

```rust
#[derive(Clone, Serialize, Deserialize, PartialEq)]
pub struct Store {
   pub id: i32,

    #[serde(
        serialize_with = "serialize_looker",
        deserialize_with = "deserialize_looker"
    )]
   pub looker: SomeCustomLooker // coming from external library
}
```

but not sure, how to tackle this with `rkvy`.

MaxOhn:
Closes #446

### DIFF
--- a/rkyv/examples/remote_types.rs
+++ b/rkyv/examples/remote_types.rs
@@ -1,0 +1,118 @@
+use rancor::Failure;
+use rkyv::{with::With, Archive, Deserialize, Serialize};
+
+// Assume this is a remote module or crate that you cannot modify.
+mod remote {
+    // Notably, this type does not implement the rkyv traits
+    #[derive(Debug, PartialEq)]
+    pub struct Foo {
+        pub ch: char,
+        pub bytes: [u8; 4],
+        pub _uninteresting: u32,
+        // ... and even has private fields
+        bar: Bar<i32>,
+    }
+
+    #[derive(Debug, PartialEq)]
+    pub struct Bar<T>(pub T);
+
+    impl Foo {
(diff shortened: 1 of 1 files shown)


rkyv/examples/remote_types.rs
```diff
@@ -0,0 +1,118 @@
...
+        }
+    }
+}
+
+// Let's create a local type that will serve as `with`-wrapper for `Foo`.
+// Fields must have the same name and type but it's not required to define all
+// fields.
+#[derive(Archive, Serialize, Deserialize)]
+#[rkyv(remote = remote::Foo)] // <-
+#[rkyv(archived = ArchivedFoo)]
```
djkoloski, line 48: I really like this idea; maybe the default name for archived types should be made from the remote type. I’ll make a quick change to add this.

djkoloski (approved):

djkoloski:
Thanks! This example is great!

PR #550, opened 2024-09-14
fix: minimum `syn` version

xJonathanLEI:
`rkyv` uses `Fields::members()` which was added in `syn` [v2.0.73](https://github.com/dtolnay/syn/releases/tag/2.0.73) (not a very correct use of SemVer on `syn`'s part). Downstream projects that have locked `syn` to an older version would fail to compile upon adding `rkyv` as a dependency.

This can be reproduced by just trying to compile:

```toml
[package]
name = "rkyv-test"
version = "0.1.0"
edition = "2021"

[dependencies]
rkyv = "0.8.2"
syn = "=2.0.72"
```


djkoloski:
Thanks for the PR, sorry about that!

PR #554, opened 2024-09-16
Add a generic niching wrapper

MaxOhn:
Supersedes #483.

Adds the with-wrapper `Nicher<D>` to allow customizable niching based on `D` which needs to implement the new trait `Decider<T>` to niche an instance of `T`.

Implemented deciders are `Zero` for the `NonZero` types, `NaN` for `f32` and `f64`, and `Null` for `Box<T>`. Additional implementations probably belong into separate PRs.

One thing about this very flexible approach is that it supports many more use-cases than are reasonable to support in rkyv. For example there's nothing inherently wrong with implementing `Decider<f32> for Zero` to niche `Some(0.0)` into `None` but it's probably fine to keep the amount of implementations manageable and let users handle these cases themselves.

### DIFF
--- a/rkyv/src/niche/decider.rs
+++ b/rkyv/src/niche/decider.rs
@@ -1,0 +1,30 @@
+//! Deciders for niching values with [`Nicher`].
+//!
+//! [`Nicher`]: crate::with::Nicher
+
+use crate::{Archive, Archived, Place, Portable};
+
+/// A type that can be used to niche a value with [`Nicher`].
+///
+/// [`Nicher`]: crate::with::Nicher
+pub trait Decider<T: Archive> {
+    /// The archived representation of both niched and non-niched values.
+    type Archived: Portable;
+
+    /// Converts a niched archive to `None`; otherwise `Some(_)`.
+    fn as_option(archived: &Self::Archived) -> Option<&Archived<T>>;
+
+    /// Creates a `Self::Archived` from an `Option<&T>` and writes it to the
+    /// given output.
+    fn resolve_from_option(
+        option: Option<&T>,
+        resolver: Option<T::Resolver>,
+        out: Place<Self::Archived>,
+    );
+}
+
+/// [`Decider`] for NaN-niched values.
+pub struct NaN;
+
+/// [`Decider`] for null-pointer-niched values.
+pub struct Null;
--- a/rkyv/src/niche/decider.rs
+++ b/rkyv/src/niche/decider.rs
@@ -50,20 +50,12 @@ use crate::{Archive, Place, Portable};
 /// # fn main() -> Result<(), rkyv::rancor::Error> {
 /// assert!(size_of::<ArchivedNiched>() < size_of::<ArchivedBasic>());
 ///
-/// # #[cfg(feature = "alloc")] {
 /// let values = [Niched(Some(1)), Niched(Some(42)), Niched(None)];
 /// let bytes = rkyv::to_bytes(&values)?;
-/// # #[cfg(feature = "bytecheck")]
-/// let mut iter = rkyv::access::<[ArchivedNiched; 3], _>(&bytes)?.iter();
-/// # #[cfg(not(feature = "bytecheck"))]
-/// # let mut iter = unsafe {
-/// #     rkyv::access_unchecked::<[ArchivedNiched; 3]>(&bytes)
-/// # }.iter();
-/// assert_eq!(iter.next().unwrap().0.as_ref(), None);
-/// assert_eq!(iter.next().unwrap().0.as_ref(), Some(&42.into()));
-/// assert_eq!(iter.next().unwrap().0.as_ref(), None);
-/// assert!(iter.next().is_none());
-/// # }
+/// let archived = rkyv::access::<[ArchivedNiched; 3], _>(&bytes)?;
+/// assert_eq!(archived[0].0.as_ref(), None);
+/// assert_eq!(archived[1].0.as_ref(), Some(&42.into()));
+/// assert_eq!(archived[2].0.as_ref(), None);
 /// # Ok(()) }
 /// ```
 ///
--- a/rkyv/src/niche/decider.rs
+++ b/rkyv/src/niche/decider.rs
@@ -2,18 +2,17 @@
 //!
 //! [`Nicher`]: crate::with::Nicher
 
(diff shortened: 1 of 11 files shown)


MaxOhn:
As for implementational details, I saw that `ArchivedOptionBox` [unwrap the resolver via `unreachable_unchecked`](<https://github.com/MaxOhn/rkyv/blob/generic-niche/rkyv/src/niche/option_box.rs#L168>) which seems reasonable but on the other hand gives users the opportunity to invoke UB through safe code so instead I [opted to panic](<https://github.com/MaxOhn/rkyv/blob/generic-niche/rkyv/src/impls/core/with/decider.rs#L26>).

rkyv/src/niche/niched_option.rs
```diff
@@ -0,0 +1,285 @@
...
+    where
+        T: Archive<Archived: CheckBytes<C>>,
+        D: Decider<T, Niched: CheckBytes<C>> + ?Sized,
+        C: Fallible + ?Sized,
+    {
+        unsafe fn check_bytes(
+            value: *const Self,
+            context: &mut C,
+        ) -> Result<(), C::Error> {
+            unsafe { <D::Niched>::check_bytes(&*(*value).niche, context)? };
```
djkoloski, line 100: There's a very minor risk that checking the niched type and then the un-niched type could cause problems if they both mutate the context. Do you think it would be useful to require `D::Niched: CheckBytes<()>`? (meant to indicate "this type can be checked without any additional context").

I could see someone wanting to access the context for benign reasons though (like "runtime niching" or something), so maybe it's just best to document that this type may check the niched type as well as the un-niched type.

rkyv/src/with.rs
```diff
@@ -473,6 +473,57 @@ pub struct AsVec;
 #[derive(Debug)]
 pub struct Niche;
 
+/// A wrapper that niches based on a generic [`Decider`].
+///
+/// A common type combination is `Option<Box<T>>`. By using a null pointer, the
```
djkoloski, line 478:
```suggestion
/// A common type combination is `Option<Box<T>>`. By niching `None` into the null pointer, the
```

rkyv/src/with.rs
```diff
@@ -473,6 +473,57 @@ pub struct AsVec;
...
+/// }
+///
+/// assert!(
+///     size_of::<Archived<BasicExample>>()
+///         > size_of::<Archived<NichedExample>>()
+/// );
+/// ```
+///
+/// [`Decider`]: crate::niche::decider::Decider
+pub struct Nicher<D: ?Sized>(PhantomData<D>);
```
djkoloski, line 513: I did some tests and, as far as I can tell, we could add a decider type parameter to the existing `Niche` wrapper _as long as_ we provide a default. It seems reasonable to add a `Default` decider which only implements `Decider` for "reasonable" types like boxes and the non-zero types. This would let us keep the ergonomics of `#[rkyv(with = Niche)]` while still allowing the flexibility of `#[rkyv(with = Niche<One>)]`.

djkoloski (changes requested):
This looks really good! I'm thoroughly pleased with this implementation. I left some comments mostly related to naming, but I'm happy with the direction.

rkyv/src/niche/decider.rs
```diff
@@ -0,0 +1,90 @@
...
+/// assert_eq!(iter.next().unwrap().0.as_ref(), Some(&42.into()));
+/// assert_eq!(iter.next().unwrap().0.as_ref(), None);
+/// assert!(iter.next().is_none());
+/// # }
+/// # Ok(()) }
+/// ```
+///
+/// [`Nicher`]: crate::with::Nicher
+/// [`is_niched`]: Decider::is_niched
+pub unsafe trait Decider<T: Archive> {
```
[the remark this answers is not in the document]
MaxOhn: Agreed

rkyv/src/niche/niched_option.rs
```diff
@@ -0,0 +1,285 @@
...
+
+impl<T, D> NichedOption<T, D>
+where
+    T: Archive,
+    D: Decider<T> + ?Sized,
+{
+    /// Returns `true` if the option is a `None` value.
+    pub fn is_none(&self) -> bool {
+        #[allow(clippy::let_unit_value)]
+        let _ = Repr::<T, D>::NICHE_SIZE_CHECK;
```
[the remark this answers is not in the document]
MaxOhn: Credits to @novacrazy 😄

rkyv/src/with.rs
MaxOhn: I was also thinking of how to combine `Niche` with the new generic version but wanted approval for this whole approach first 😄 

It would definitely be possible to add a `DefaultNiching` to replace the two nichings that were already supported (`Box<T>` and `NonZero*`) and then turn the `Niche` struct into a type alias for `Nicher<DefaultNiching>`. The small downside is that `Niche` would no longer be a unit struct so that's technically a breaking change. The bigger downside is that the niched archived types would be replaced with the new `NichedOption` type which is most definitely a breaking change.

I think the better solution would be to mark `Niche` as deprecated with an info about `Nicher` and then remove it entirely for the next breaking release.

rkyv/src/niche/niched_option.rs
MaxOhn: True, good catch.

I think it might just be sufficient to check the niched type and be done with it.
- If it fails we return an error
- If it succeeds and `D::is_niched` would return true then we return ok.
- If it succeeds and `D::is_niched` would return false then the safety invariant of `Niching` as it is currently stated requires the implementor to ensure that `T` must then be safe to access.

Does that sound right?

novacrazy: I agree, though it should be written down in the docs that the niched type should also `check_bytes` on any inner non-niched values, directly or indirectly. Like for `ArchivedBox<T>`, the niched repr is the same as the internal of `ArchivedBox` anyway, so double-checking is unnecessary. Same with the non-zeros and floats.

rkyv/src/with.rs
djkoloski: Ah, that's a great point. I agree with your suggestion, let's land this without changing anything about `Niche`. I'll deprecate it before the next point release.

djkoloski (approved):

PR #560, opened 2024-10-09
Upgrade to hashbrown 0.15

novacrazy:
This was very simple thanks to the versioned feature flags we added. I also cleaned up the code to make copy-pasting even easier in the future.


djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #580, opened 2024-12-23
docs: Fix incorrect URL for Climatiq Update README.md

0xwitty:
I noticed a issue in the "Thanks" section of the README file. The URL for Climatiq was pointing to an incorrect domain (`https://climatiq.hq`). I’ve updated it to the correct one: `https://www.climatiq.io`.

This should ensure the link directs users to the right website. Thanks!


djkoloski:
Thanks for the PR!

Issue #596, opened 2025-02-24
`#![deny(missing_docs)]` fails on derived archived enum variants


tv42:
This looks like #217 again. It seems ArchivedFoo enum variants don't have doc strings. Either they could get docstrings, or `#[expect(missing_docs, reason = "generated code")]` or something.

```rust
//! Repro for an rkyv bug

#![deny(missing_docs)]

/// doc for enum
#[derive(rkyv::Archive)]
pub enum A {
    /// doc for variant
    Variant {
        /// doc for field
        field: (),
    },
}

/// doc for enum
#[derive(rkyv::Archive)]
// Can't seem to special case allow it, this doesn't apply to enum variants?
#[rkyv(attr(expect(missing_docs)))]
pub enum B {
    /// doc for variant
    Variant {
        /// doc for field
        field: (),
    },
}

/// doc for enum
#[derive(rkyv::Archive)]
#[rkyv(derive(Debug), attr(expect(missing_docs)))]
pub enum C {
    /// doc for variant
    // Can't special case allow it this way either.
    #[rkyv(attr(expect(missing_docs)))]
    Variant {
        /// doc for field
        field: (),
    },
}
```

```console
$ cargo clippy
    Checking rkyv-missing-docs-issue v0.1.0 (/workspace/rkyv-missing-docs-issue)
error: missing documentation for a struct field
 --> src/lib.rs:6:10
  |
6 | #[derive(rkyv::Archive)]
  |          ^^^^^^^^^^^^^
  |
note: the lint level is defined here
 --> src/lib.rs:3:9
  |
3 | #![deny(missing_docs)]
  |         ^^^^^^^^^^^^
  = note: this error originates in the derive macro `rkyv::Archive` (in Nightly builds, run with -Z macro-backtrace for more info)

error: missing documentation for a struct field
 --> src/lib.rs:6:10
  |
6 | #[derive(rkyv::Archive)]
  |          ^^^^^^^^^^^^^
  |
  = note: this error originates in the derive macro `rkyv::Archive` (in Nightly builds, run with -Z macro-backtrace for more info)

error: missing documentation for a struct field
  --> src/lib.rs:16:10
   |
16 | #[derive(rkyv::Archive)]
   |          ^^^^^^^^^^^^^
   |
   = note: this error originates in the derive macro `rkyv::Archive` (in Nightly builds, run with -Z macro-backtrace for more info)

error: missing documentation for a struct field
  --> src/lib.rs:28:10
   |
[...]

PR #603, opened 2025-04-02
Add documentation for struct fields

Vrixyz:
Related to #596

## Tests

Run this branch with `cargo expand ArchivedTest -p rkyv --example readme` to find the added documentation.

An automated test with `deny(missing_docs)` would be a great addition but let's discuss first.

### DIFF
--- a/rkyv_derive/src/archive/struct.rs
+++ b/rkyv_derive/src/archive/struct.rs
@@ -176,11 +176,18 @@ fn generate_archived_type(
             ..
         } = field;
 
+        let field_name = ident.as_ref();
+        let field_doc = format!(
+            "The archived counterpart of [`{}:{}`]",
+            name,
+            field_name.map_or("unnamed".to_string(), |s| s.to_string())
+        );
         let field_attrs = FieldAttributes::parse(attributes, field)?;
         let field_metas = field_attrs.metas();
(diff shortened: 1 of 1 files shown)


rkyv_derive/src/archive/struct.rs
```diff
@@ -176,11 +176,18 @@ fn generate_archived_type(
             ..
         } = field;
 
+        let field_name = ident.as_ref();
+        let field_doc = format!(
+            "The archived counterpart of [`{}:{}`]",
+            name,
+            field_name.map_or("unnamed".to_string(), |s| s.to_string())
```
Vrixyz, line 183: unnamed fields should probably be better handled with their indices, but I'll continue after a validation.

djkoloski (approved):

djkoloski (changes requested):

rkyv_derive/src/archive/struct.rs
Vrixyz: Thanks! Should we have testing on those documentations ?

It's a tiny bit tricky to test because we should probably create a new library dedicated to testing which should deny missing docs 🤔 ; maybe better to leave for another PR?

djkoloski (approved):

PR #631, opened 2025-10-19
Implement raw entry API for hashmap

anchpop:
Hi. I needed this because certain methods of interning strings can be made much more memory efficient if you have the ability to "fake" hashes in a hashmap. I learned this trick from the library `lasso`. 

By the way, I'm not the best at unsafe code, so please forgive me if something is wrong there.

### DIFF
--- a/rkyv/src/collections/swiss_table/map.rs
+++ b/rkyv/src/collections/swiss_table/map.rs
@@ -90,6 +90,14 @@ impl<K, V, H> ArchivedHashMap<K, V, H> {
             _phantom: PhantomData,
         }
     }
+
+    /// Creates a raw entry builder for low-level access to entries.
+    ///
+    /// This allows accessing entries by hash and custom comparison functions,
+    /// similar to `hashbrown::HashMap::raw_entry()`.
+    pub fn raw_entry(&self) -> RawEntryBuilder<'_, K, V, H> {
+        RawEntryBuilder { map: self }
+    }
 }
 
 impl<K, V, H: Hasher + Default> ArchivedHashMap<K, V, H> {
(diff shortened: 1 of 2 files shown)


rkyv/src/collections/swiss_table/table.rs
djkoloski: The changes in this file don't appear to add any functionality. `get_with` is redundant with `get_entry_raw`, so I would prefer to leave the naming unchanged unless there's a really compelling reason to update it.

rkyv/src/collections/swiss_table/map.rs
```diff
@@ -419,3 +454,43 @@ impl<K, V, H> ExactSizeIterator for ValuesMut<'_, K, V, H> {
 }
 
 impl<K, V, H> FusedIterator for ValuesMut<'_, K, V, H> {}
+
+/// A builder for accessing entries in an [`ArchivedHashMap`] with a custom hash and comparison function.
+///
+/// This is a lower-level API similar to `hashbrown::HashMap::raw_entry()`.
+pub struct RawEntryBuilder<'a, K, V, H> {
```
djkoloski: This is a large API surface, and rkyv's goal is to match (as closely as possible) the API for items in `std`, not `hashbrown`.

anchpop:
Yeah, the user of the code in this PR has to make sure they use `let hash = rkyv::hash::hash_value::<str, rkyv::hash::FxHasher64>(val);` (or something like that) to generate the hashes. As you mention, it would not be advisable for someone to use their own hash function, unless they were sure it was platform-agnostic like rkyv's.

I didn't realize that the goal was to match as closely as possible the API in `std`. In that case, I think this use case is probably too niche to justify deviating from that principle. (Especially because of the potential for user error it introduces.) So I'll close the PR for now. If I can think of an API design that's simpler and harder to misuse but still allows the optimization for string interning, I'll leave a comment.

PR #635, opened 2025-11-25
Provide alignment of bytes written to `Writer`.

victoryaskevich:
Hi!

When serializing rkyv data it seems there is no reliable way to know the required alignment of serialized bytes. As I understand currently it works by assuming that alignment 16 is enough for practical cases. This can be both not enough, and too much.

In my case, I actually pack many independent rkyved bytes together, in a single buffer, so I want to know the minimal required alignment, which can be <16 in many cases, to save space on unneeded padding. That is why I want to know the exact requirement, not large enough assumption.

I added `align` parameter to `Writer::write` method, so for each write the intended alignment of written value is communicated, and custom implementation of `Writer` can use it.

Also I added `write_padding` method directly to `Writer` trait, so it can be overwritten with more efficient implementation.

I am marking it as "draft" PR to discuss this issue and possible solution.

### DIFF
--- a/rkyv/src/collections/btree/map/mod.rs
+++ b/rkyv/src/collections/btree/map/mod.rs
@@ -575,7 +575,7 @@ impl<K, V, const E: usize> ArchivedBTreeMap<K, V, E> {
                 size_of::<LeafNode<K, V, E>>(),
             )
         };
-        serializer.write(bytes)?;
+        serializer.write(align_of::<LeafNode<K, V, E>>(), bytes)?;
 
         Ok(pos)
     }
@@ -654,7 +654,7 @@ impl<K, V, const E: usize> ArchivedBTreeMap<K, V, E> {
                 size_of::<InnerNode<K, V, E>>(),
             )
         };
-        serializer.write(bytes)?;
+        serializer.write(align_of::<InnerNode<K, V, E>>(), bytes)?;
 
         Ok(pos)
     }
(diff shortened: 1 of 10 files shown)


victoryaskevich:
I though of a different workaround, where I run validation multiple times on buffers with various alignments, so I can probe the minimal required alignment for the data. But I think this should be solved directly in `rkyv`, so I decided to open this topic before it hits `v1.0`, and no breaking changes are accepted

soruh:
By checking only 0 and 1 you also get the minimal alignment, but yeah, I agree that it seems important for `rkyv` to provide tools to probe how much alignment is needed, given how important it is for preventing UB.

Although, from what I've seen, would it maybe be possible to hook into the check logic, given that it can already check that the alignment is correct?

djkoloski:
@soruh I think you should be able to just write directly to the file by creating a custom serializer. [`IoWriter::with_pos`](https://docs.rs/rkyv/latest/rkyv/ser/writer/struct.IoWriter.html#method.with_pos) serves a similar purpose, since you may want to write bytes to an output which already contains bytes. You'd probably just want a similar idea but for byte buffers (or maybe `IoWriter` fits your use case).

victoryaskevich:
Re-purposing validation for this looks the most fragile. Is it free from early returns, and guarantees everything is visited. Does this property hold with custom validation for custom types (I have custom validators). And my custom validators aren’t really supposed to be created at serialization step, I don’t have enough context for this. I will need even more fragile workarounds.

Issue #628, opened 2025-09-23
Hashbrown 0.16 support?


VorpalBlade:
Hashbrown 0.16 is out. It would be good to have support in an rkyv release.

PR #638, opened 2025-12-01
Add support for hashbrown 0.16

dmackinn:
Addresses #628

- Updated workspace dependency to hashbrown 0.16
- Created new hashbrown_0_16 module with HashMap/HashSet implementations
- Updated hashbrown_0_15 to use separate package dependency
- Added hashbrown-0_16 feature flag

Confirmed that the tests still pass.


dmackinn:
> Since we're now up to three separate versions of hashbrown which are supported, I think it would be best to unify as much of these impls as possible. Could you unify these impls via macro to add hashbrown 0.16 support?

Sounds good!

Issue #644, opened 2025-12-30
Potential Undefined Behaviors in `Arc<T>/Rc<T>` impls of `from_value` on OOM


shinmao:
The `SharedPointer::alloc` implementation for `sync::Arc<T>` in `rkyv/src/impls/alloc/rc/atomic.rs` does not check if the allocator returns a null pointer on OOM (The possibility of [memory is exhausted](https://doc.rust-lang.org/alloc/alloc/trait.GlobalAlloc.html#tymethod.alloc) in mentioned in the documentation). This null pointer can flow through to `SharedPointer::from_value`, which calls `Box::from_raw(ptr)` with a null pointer, triggering **undefined behavior** via safe deserialization APIs.

https://github.com/rkyv/rkyv/blob/2c49abc663699c9513199c9c419e32fd4674d27f/rkyv/src/impls/alloc/rc/atomic.rs#L44-L58

* Line 48 can return null on OOM.
* Line 53 wraps null pointer in `Ok`.
* Line 57 triggers undefined behavior when `ptr` is null.

## Trigger Path (Safe API)
The bug is reachable through **safe public APIs**:
```
from_bytes → deserialize_using -> Deserialize::deserialize → deserialize_shared 
```
Following is the code of `deserialize_shared`:
https://github.com/rkyv/rkyv/blob/2c49abc663699c9513199c9c419e32fd4674d27f/rkyv/src/de/pooling/mod.rs#L257-L261
```
-> SharedPointer::alloc → SharedPointer::from_value → Box::from_raw(null)
```
This can totally be triggered within safe-only Rust code, which violates the safety promise of Rust.

shinmao:
At the same time, `Rc<T>` also shows the same pattern with `alloc`,
https://github.com/rkyv/rkyv/blob/2c49abc663699c9513199c9c419e32fd4674d27f/rkyv/src/impls/alloc/rc/mod.rs#L47-L61
which could also trigger undefined behaviors in `from_value`.

shinmao:
@djkoloski thanks for the fast response. After you release the next version, do you agree with me to publish the rustsec advisory for you?

djc:
> You can publish whatever you'd like in the rustsec database, I don't really put much stock in it.

Out of curiosity, why's that?

shinmao:
Hi @djkoloski , thanks for the fast fix and for sharing your perspective.

[...]

PR #645, opened 2026-01-05
Backport null pointer alloc check from b48a86e (#644)

mcronce:
Think I got everything.  Let me know what doesn't look good


djkoloski (approved):

mcronce:
I don't _think_ that `miri` failure is related - if I'm reading correctly, it's in `rkyv_dyn`, but I only modified `rkyv` and it doesn't seem to be happening anywhere that it calls into `rkyv_dyn` calls into `rkyv`.  That said, the way those macros interact is pretty opaque to me without looking deeper; if you think I'm not correct, I'll take a closer look.

PR #648, opened 2026-01-27
Remove unnecesssary bounds

anchpop:
Removing these bounds saved me >300 lines in my project. Types that do not implement the bounds will be of limited usefulness when archived, but if you don't use the archived versions of the types except to deserialize them it's nice not to have to specify the bounds.


djkoloski:
Sorry for the delay in reviewing. I'm not sure if this change would affect any entailed bounds, and if it would then this would be a breaking change. Do you know whether that's the case? This may be a case where running cargo-semver-checks would be helpful for the project.

PR #652, opened 2026-02-20
Add `#[repr(C)]` for `AlignedVec`

msrd0:
When going Rust to Rust via the C ABI, we need FFI safe types. `AlignedVec`'s fields are already perfectly suitable for an FFI boundary, so adding `#[repr(C)]` is the only change necessary to allow that to happen safely (if you can ever call FFI safe ...).


djkoloski:
I'd prefer not to make guarantees about the FFI-safety of rkyv's types. Would `from_raw_parts`/`into_raw_parts` fit your use case? Then you could cross the FFI boundary safely by passing the individual parts of the `AlignedVec`.

msrd0:
Yes that works for me, closing in favour of #653

PR #653, opened 2026-02-20
Add `into_parts` and `from_parts` for `AlignedVec`

msrd0:
These methods are helpful if you need to pass an `AlignedVec` across an FFI boundary, as the `AlignedVec` type itself doesn't have a stable ABI.

I took some inspiration, both regarding the naming and the documentation, from [`Vec::into_parts`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.into_parts) and [`Vec::from_parts`](https://doc.rust-lang.org/stable/std/vec/struct.Vec.html#method.from_parts).

### DIFF
--- a/rkyv/src/util/alloc/aligned_vec.rs
+++ b/rkyv/src/util/alloc/aligned_vec.rs
@@ -6,6 +6,7 @@ use core::{
     ptr::NonNull,
     slice,
 };
+use std::mem::ManuallyDrop;
 
 use rancor::Fallible;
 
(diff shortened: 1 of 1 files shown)


rkyv/src/util/alloc/aligned_vec.rs
```diff
@@ -6,6 +6,7 @@ use core::{
     ptr::NonNull,
     slice,
 };
+use std::mem::ManuallyDrop;
```
djkoloski: Please move this into the `core` use block above (i.e. `core::{mem::ManuallyDrop}`). rkyv supports `alloc` without `std` so we have to prefer `core` whenever possible.

djkoloski (commented):
One small change, otherwise LGTM

rkyv/src/util/alloc/aligned_vec.rs
msrd0: Ah yes, rust-analyzer's autocompletion does a bad job at std vs core ...

djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #655, opened 2026-03-26
Raw bytes support for ArchivedStringRepr

MattWhelan:
This struct is a useful building block for interning non-standard string representations, but using `&str` parameters sharply limits that utility, in that it forces a byte buffer to be coerced into a `str` only to get its length.

This change offers `raw` variants of the `emplace_` methods, which take byte slices rather than `&str`, allowing a user to avoid potential UB.

### DIFF
--- a/rkyv/src/string/repr.rs
+++ b/rkyv/src/string/repr.rs
@@ -158,7 +158,11 @@ impl ArchivedStringRepr {
         Seal::new(unsafe { str::from_utf8_unchecked_mut(bytes) })
     }
 
-    /// Emplaces a new inline representation for the given `str`.
+    /// Emplaces a new inline representation for the given byte slice.
+    ///
+    /// This is the raw version of [`emplace_inline`](Self::emplace_inline)
+    /// that accepts `&[u8]` instead of `&str`, making it suitable for use
+    /// with non-UTF-8 data.
     ///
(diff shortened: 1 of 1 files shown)


djkoloski:
I took a quick glance at this PR. It appears to be unsound because `ArchivedStringRepr` relies on the first byte of an inlined string being a valid first byte for UTF-8 (see: `ArchivedStringRepr::is_inline` checks whether `self.inline.bytes[0] & 0xc0 != 0x80`). If your intention is to support byte buffers that are known to be UTF-8, but without calling `str::from_utf8`, then you should call `str::from_utf8_unchecked` and then use the resulting `str` with `ArchivedStringRepr`.

MattWhelan:
Thanks for the quick response! I missed that.

What I had in mind was non-UTF-8 string representations, which is a bit problematic given that discriminator strategy.

djkoloski:
That's unfortunate, sorry to hear that. Inlining small strings is relatively easy since we know that they're just bytes. By comparison, inlining small vecs or slices (like `ArchivedVec<u8>`) is much harder because it adds a lot of complexity across a very generic API surface. Maybe not impossible, but gets much farther into issues related to computing constants that use generic parameters which is not currently supported in Rust.

PR #656, opened 2026-04-01
feat: implement `NoUndef` for `[T]` where `T: NoUndef`.

Ddystopia:
It is similar to `unsafe impl NoUndef for str {}` and `unsafe impl<T: Portable> Portable for [T] {}` that are already present in the codebase.


djkoloski (approved):

djkoloski:
Thanks for the PR!

PR #660, opened 2026-04-18
Link to the examples directory, instead of away from examples

cceckman:
When trying to figure out how to use rkyv, I got stuck in a loop.

https://rkyv.org/ says the book "is the best way to learn and understand rkyv, but won’t go as in-depth on specifics as the documentation will." While the book had a lot of explanation of how rkyv works under the hood, I did not find any full examples of "here is how to archive end-to-end".

https://docs.rs/rkyv/latest/rkyv/ does not contain any examples at the top level either, nor links to any. It says "See the rkyv book for guide-level documentation and usage examples."

Fix this: point to where the examples actually are, in the source tree.

### DIFF
--- a/rkyv/src/lib.rs
+++ b/rkyv/src/lib.rs
@@ -25,10 +25,10 @@
 //! rkyv's types will have exactly the same performance as native types.
 //!
 //! See the [rkyv book] for guide-level documentation.
-//! See [the examples directory] for usage examples.
+//! See [the examples directory][1] for usage examples.
 //!
 //! [rkyv book]: https://rkyv.org
-//! [the examples directory]: https://github.com/rkyv/rkyv/tree/main/rkyv/examples
+//! [1]: https://github.com/rkyv/rkyv/tree/main/rkyv/examples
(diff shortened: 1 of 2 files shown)


rkyv/src/lib.rs
```diff
@@ -24,9 +24,11 @@
 //! zero-copy types are designed to have little to no overhead. In most cases,
 //! rkyv's types will have exactly the same performance as native types.
 //!
-//! See the [rkyv book] for guide-level documentation and usage examples.
+//! See the [rkyv book] for guide-level documentation.
+//! See [the examples directory] for usage examples.
 //!
 //! [rkyv book]: https://rkyv.org
+//! [the source]: https://github.com/rkyv/rkyv/tree/main/rkyv/examples
```
djkoloski, line 31:
```suggestion
//! [the examples directory]: https://github.com/rkyv/rkyv/tree/main/rkyv/examples
```

djkoloski (changes requested):

PR #673, opened 2026-05-25
Fix Rc deserialization soundness issue

### LINKED ISSUE #670: Out-of-bounds reads in deserializing crafted archives with `Rc`
Insufficient validation in `ArchivedRc::verify` causes out-of-bounds reads on deserializing crafted archives.

## Steps to reproduce

Add the following test to rkyv/src/rc.rs and run `MIRIFLAGS=-Zmiri-tree-borrows cargo +nightly miri test --release --features=alloc,bytecheck shared_unsized_metadata_can_change_after_first_validation`

```rust
#[cfg(all(test, miri, feature = "alloc", feature = "bytecheck"))]
mod tests {
    use crate::{
        access,
        alloc::{rc::Rc, vec},
        to_bytes, Archive, Serialize,
    };

    #[test]
    fn shared_unsized_metadata_can_change_after_first_validation() {
        #[derive(Archive, Serialize)]
        #[rkyv(crate)]
        struct Test {
            good: Rc<[u32]>,
            bad: Rc<[u32]>,
        }

        // Shared pointer validation keys already-validated pointees by data
        // address and type only. For unsized pointees, metadata is part of the
        // pointer validity invariant too. After `good` validates the shared
        // slice with length 1, `bad` can reuse the same data address with a
        // forged length and skip validating the larger slice. Safe indexing
        // then trusts the forged metadata and reads outside the allocation.
        let shared = Rc::<[u32]>::from(vec![123].into_boxed_slice());
        let value = Test {
            good: shared.clone(),
            bad: shared,
        };
        let mut bytes = to_bytes::<rancor::Error>(&value).unwrap();

        let bad_metadata_offset = {
            let archived =
                access::<ArchivedTest, rancor::Error>(&bytes).unwrap();
            let metadata = archived.bad.ptr.metadata();
            metadata as *const _ as usize - bytes.as_ptr() as usize
        };

        let forged_len_bytes = to_bytes::<rancor::Error>(&1024usize).unwrap();
        bytes
            [bad_metadata_offset..bad_metadata_offset + forged_len_bytes.len()]
            .copy_from_slice(&forged_len_bytes);

        let archived = access::<ArchivedTest, rancor::Error>(&bytes).unwrap();
        let _ = archived.bad[100].to_native();
    }
}
```

## Impact

This is **not just API misuse**. It is a checked-validation soundness issue for untrusted archives, and it can flow into checked deserialization.

The core bug is here: [rkyv/src/rc.rs#L356](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/rc.rs#L356). `ArchivedRc::verify` keys shared validation by only the pointee data address plus `TypeId`. `SharedValidator` stores only `usize -> (TypeId, bool)` at [validator.rs#L20](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/validation/shared/validator.rs#L20). For DSTs, the fat-pointer metadata is part of validity, but when the second `Rc<[u32]>` points to the same data address with a forged length, validation returns `Finished` and skips the bounds/type check at [rc.rs#L369](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/rc.rs#L369).

It **can be triggered through checked deserialization**. `from_bytes` first calls checked `access` at [high/checked.rs#L249](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/api/high/checked.rs#L249), so the same validation bypass lets the malformed archive through. Then `ArchivedRc::deserialize` calls `self.get()` at [impls/alloc/rc/mod.rs#L87](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/impls/alloc/rc/mod.rs#L87), creating/using a forged fat reference. The pooling layer also keys by archived data address only and rebuilds a returned pointer using the later metadata at [de/pooling/mod.rs#L254](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/de/pooling/mod.rs#L254) and [de/pooling/mod.rs#L274](https://github.com/rkyv/rkyv/blob/4686a550931fd28176988b26e5ab71731ac9decf/rkyv/src/de/pooling/mod.rs#L274). So a checked `from_bytes` variant of the PoC should be possible by deriving `Deserialize` and using/allowing drop of the resulting `bad: Rc<[u32]>`.

For **ordinary safe serialization of `Rc<[T]>`/`Arc<[T]>`**, I do not see this exact issue being produced. `to_bytes` of two clones writes the same archived metadata for both pointers; the PoC needs tampered bytes or an invalid/unsafe/custom DST setup to create same-address, different-metadata shared pointers. The serializer also keys sharing by data address only, so custom DSTs or trait-object cases may deserve a separate look, but the concrete slice PoC is primarily a validation/deserialization-of-malicious-bytes bug.

Suggested severity: **soundness/security bug in checked access/deserialization for untrusted data**, affecting shared pointers to unsized pointees (`Rc`, `Arc`, and likely `Weak` validation paths). Not merely misuse, but not a normal safe-serialization-originated bug for standard `Rc<[T]>` values.

## Credits

The issue was discovered by GPT-5.5. The PoC is LLM-authored and verified manually before reporting. The impact analysis is LLM-authored.

Shnatsel:
Fixes #670

The first commit adds the UB PoC from #670, the second commit fixes the UB loophole. The details are in the commit message.

This is authored by GPT-5.5 and human-reviewed. I decided to see if an LLM that detected the issue can also fix it, and apparently it can. Please let me know if this kind of PR is helpful or not.

A wrinkle that LLMs did _not_ point out: `TypeId` is a hash and theoretically can collide. But the values are not attacker-controlled and we only have a handful of types that we generate a TypeId (the smart pointer types like Rc, Arc and Weak plus user-defined extensions) so I don't think it's an issue in practice.

### DIFF
--- a/rkyv/src/rc.rs
+++ b/rkyv/src/rc.rs
@@ -352,14 +352,15 @@ mod verify {
         fn verify(&self, context: &mut C) -> Result<(), C::Error> {
             let ptr = self.ptr.as_ptr_wrapping();
             let type_id = TypeId::of::<ArchivedRc<T, F>>();
+            let metadata = self.ptr.metadata();
 
             let addr = ptr as *const u8 as usize;
-            match context.start_shared(addr, type_id)? {
+            match context.start_shared(addr, type_id, metadata)? {
                 ValidationState::Started => {
                     context.in_subtree(ptr, |context| unsafe {
                         T::check_bytes(ptr, context)
                     })?;
-                    context.finish_shared(addr, type_id)?;
+                    context.finish_shared(addr, type_id, metadata)?;
                 }
                 ValidationState::Pending => {
                     if !F::ALLOW_CYCLES {
@@ -397,29 +398,22 @@ mod verify {
     }
 }
 
-#[cfg(all(test, miri, feature = "alloc", feature = "bytecheck"))]
+#[cfg(all(test, feature = "alloc", feature = "bytecheck"))]
 mod tests {
     use crate::{
         access,
         alloc::{rc::Rc, vec},
-        to_bytes, Archive, Serialize,
+        from_bytes, to_bytes, Archive, Deserialize, Serialize,
     };
 
-    #[test]
-    fn shared_unsized_metadata_can_change_after_first_validation() {
-        #[derive(Archive, Serialize)]
-        #[rkyv(crate)]
-        struct Test {
-            good: Rc<[u32]>,
-            bad: Rc<[u32]>,
-        }
+    #[derive(Archive, Deserialize, Serialize)]
+    #[rkyv(crate)]
+    struct Test {
+        good: Rc<[u32]>,
+        bad: Rc<[u32]>,
+    }
 
-        // Shared pointer validation keys already-validated pointees by data
-        // address and type only. For unsized pointees, metadata is part of the
-        // pointer validity invariant too. After `good` validates the shared
-        // slice with length 1, `bad` can reuse the same data address with a
-        // forged length and skip validating the larger slice. Safe indexing
-        // then trusts the forged metadata and reads outside the allocation.
+    fn mismatched_shared_metadata_bytes() -> crate::util::AlignedVec {
         let shared = Rc::<[u32]>::from(vec![123].into_boxed_slice());
         let value = Test {
             good: shared.clone(),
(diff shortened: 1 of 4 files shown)


Shnatsel:
The issue was fixed in 6e13e78b4de234b367ca30160e6b9b773b80b334 instead, closing.

PR #677, opened 2026-06-09
Add `bytemuck` and `zerocopy` feature passthrough to rend

AlJohri:
Mirrors the existing `bytecheck` passthrough: rend already exposes `bytemuck::Pod`/`Zeroable` and `zerocopy::FromBytes`/`IntoBytes`/`Immutable`/`KnownLayout` for its archived primitive types behind its own `bytemuck-1` / `zerocopy-0_8` features, but downstream crates (e.g. `half`) currently have to take a direct dep on `rend` to activate them. Now they can just enable `rkyv/bytemuck` or `rkyv/zerocopy`.

Concrete motivation: needed by VoidStarKat/half-rs#150 to derive zerocopy traits on the rkyv-generated `Archivedf16` / `Archivedbf16` types without half taking a direct dep on rend.

### DIFF
--- a/rkyv/Cargo.toml
+++ b/rkyv/Cargo.toml
@@ -59,6 +59,8 @@ pointer_width_64 = []
 alloc = ["dep:hashbrown", "tinyvec-1?/alloc", "rancor/alloc"]
 std = ["alloc", "bytes-1?/std", "indexmap-2?/std", "ptr_meta/std", "uuid-1?/std"]
 bytecheck = ["dep:bytecheck", "rend/bytecheck", "rkyv_derive/bytecheck"]
+bytemuck = ["rend/bytemuck-1"]
+zerocopy = ["rend/zerocopy-0_8"]
 
 # External crate support
 hashbrown-0_17 = ["dep:hashbrown"]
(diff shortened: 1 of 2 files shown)


djkoloski:
I'd prefer to keep the zerocopy/bytemuck features of rend isolated to just rend. Maybe a workable solution here would be something like this to `half-rs`:

```toml
[features]
rkyv = ["dep:rkyv", "rend"]
bytemuck = ["dep:bytemuck", "rend?/bytemuck-1"]

[dependencies]
rend = { version = "*", optional = true, features = ["zerocopy-0_8"] }
```

which should allow you to write:

```rs
#[cfg_attr(
    all(feature = "rkyv", feature = "bytemuck"),
    rkyv(derive(bytemuck::Zeroable),
)]
#[cfg_attr(
    feature = "rkyv",
    rkyv(derive(
        zerocopy::FromBytes,
        zerocopy::IntoBytes,
        zerocopy::Immutable,
        zerocopy::KnownLayout,
    ))
)]
```

I haven't tested this to make sure it behaves exactly the way I want it to, but I think it uses all of the new cargo features in the right ways.

AlJohri:
Thanks @djkoloski! Closing this PR and moving forward in VoidStarKat/half-rs#150 with the approach you described above.

PR #681, opened 2026-07-02
Demonstrate and fix type confusion issue in Rc/Arc pooling

Shnatsel:
The first commit adds proof-of-concept unit tests that trigger UB under miri and use-after-free under Address Sanitizer due to a type confusion issue. It affects Rc and Arc but is distinct from #670

It can be reproduced with:

`RUSTFLAGS='-Zsanitizer=address' ASAN_OPTIONS=detect_leaks=0 cargo +nightly test -p rkyv --test rc_shared_type_mismatch --target=x86_64-unknown-linux-gnu`

`cargo +nightly miri test -p rkyv --test rc_shared_type_mismatch`

The detailed explanation of the issue is in the comments on the PoC tests.

The issue was discovered by GPT-5.5 so I'm reporting it publicly as per the security policy.

------

The second commit proposes a fix for the issue. A discussion of other possible approaches to fixing it is in the commit message. The fix is also generated by GPT-5.5, please do with it as you see fit.


Shnatsel:
CI failures should be fixed now.

PR #702, opened 2026-09-04
Make `LayoutRaw` an unsafe trait

### LINKED ISSUE #692: Soundness: safe `LayoutRaw` impl can underreport a DST and make access perform out-of-allocation pointer arithmetic
*Authored by an AI agent acting on joshlf's behalf.*
*This finding was identified using the [unsafe-rust skill](https://github.com/google/zerocopy/tree/main/skills/unsafe-rust).*

## Summary

`LayoutRaw` is a public, safe, unsealed trait, but validation treats its return value as proof that the complete pointee is within the archive. A downstream crate can implement `LayoutRaw` entirely in safe Rust and return a layout that is smaller than the DST described by the supplied metadata. Safe `rkyv::access` then passes a pointer to insufficient storage to `CheckBytes`. For a trailing slice, bytecheck eventually performs out-of-allocation `ptr::add`, which is undefined behavior.

This is reachable from an all-safe caller. The reproducer below has `#![forbid(unsafe_code)]`; its only deliberately false component is the safe `LayoutRaw` implementation.

Suggested severity: **high**. A downstream safe implementation plus untrusted archive bytes can make a safe validation API execute undefined behavior.

## Affected versions and current `main`

I compile-checked the witness against the published `rkyv 0.8.18` release at [`8db673416b5489f39793e654145ceb938d652c84`](https://github.com/rkyv/rkyv/commit/8db673416b5489f39793e654145ceb938d652c84). Source review found the same vulnerable safe-trait/range-check/box-validation sequence in every stable `0.8` tag from `0.8.0` through `0.8.18`. I did not replay the compiled witness on every tag and make no claim here about prereleases or `0.7.x`.

I also compile-checked the source against the current `main` head as of 2026-09-04, [`fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7`](https://github.com/rkyv/rkyv/commit/fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7). The three relevant rkyv files are unchanged between the `0.8.18` release commit and that head.

At that exact `main` revision:

- [`rkyv::access` is a safe checked-access entry point](https://github.com/rkyv/rkyv/blob/fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7/rkyv/src/api/high/checked.rs#L110-L116).
- [`LayoutRaw` is safe and has no implementer safety contract](https://github.com/rkyv/rkyv/blob/fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7/rkyv/src/traits.rs#L43-L52).
- [`ArchivedBox<T>`'s `Verify` implementation constructs the DST pointer and passes it to `in_subtree`](https://github.com/rkyv/rkyv/blob/fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7/rkyv/src/boxed.rs#L165-L190).
- [`in_subtree` trusts `T::layout_raw(metadata)` as the complete layout used by the bounds check](https://github.com/rkyv/rkyv/blob/fe0b6c04ef971ec9e633e6126f5f29d3796b9fe7/rkyv/src/validation/archive/mod.rs#L112-L147).

## Required proposition and counterexample

The unsafe validation chain requires this proposition:

> For pointer metadata `m`, `T::layout_raw(m)` either fails or returns the complete `Layout` of the `T` value described by `m`.

For the reproducer's `Underreported { head: u8, tail: [u8] }` and metadata `1024`, the complete value occupies 1025 bytes with alignment 1. Its safe `LayoutRaw` implementation instead returns `Layout { size: 1, align: 1 }`. That is a valid `Layout` value, but it falsifies the proposition consumed by validation.

The method-level prose saying that `layout_raw` returns the type's layout is not enough to make a safe trait implementation a soundness boundary: safe code is allowed to implement a safe trait without accepting any compiler-enforced unsafe obligation, and unsafe library code must remain sound for such an implementation.

## Exact safe route and UB

Under rkyv's default 0.8.18 format on the checked target (aligned, little-endian, 32-bit archived relative pointers), the twelve-byte aligned buffer below contains an eight-byte `ArchivedBox<Underreported>` root at bytes 4..12:

1. Its signed relative offset is `-4`, so the DST data address is byte 0.
2. Its archived metadata is `1024`, so the trailing `[u8]` has length 1024 and starts at byte 1.
3. Root validation accepts bytes 4..12. `ArchivedBox::verify` constructs the fat raw pointer with wrapping pointer arithmetic.
4. `in_subtree` calls the attacker-controlled safe `layout_raw(1024)`. The returned one-byte layout makes its only target-range check accept bytes 0..1.
5. Generated `CheckBytes` code reaches bytecheck's slice validator. In bytecheck 0.8.2, that validator loops over the claimed length and calls [`base.add(index)`](https://github.com/rkyv/bytecheck/blob/f97d5726d7b0cdd160f1b2de2cd3b4b60997d5a3/bytecheck/src/lib.rs#L697-L719). The built-in `u8` validator returns `Ok(())` without dereferencing, so every earlier iteration continues.
6. With `base` at allocation byte 1, indices 0..=10 remain in the allocation, index 11 produces the permitted one-past pointer, and index 12 attempts to produce allocation byte 13. The standard-library contract for [`pointer::add`](https://doc.rust-lang.org/1.93.1/std/primitive.pointer.html#method.add) states that violating its in-allocation range requirement is immediate undefined behavior. No dereference of the out-of-bounds pointer is needed.

## Compile-only reproducer — do not execute the UB function

I only ran `cargo check`; I never called `all_safe_ub_route` and never executed the undefined-behavior path. `main` is intentionally empty.

`Cargo.toml`:

```toml
[package]
name = "rkyv-layout-raw-underreport-reproducer"
version = "0.0.0"
edition = "2021"
publish = false

[[bin]]
name = "rkyv-layout-raw-underreport-reproducer"
path = "reproducer.rs"

[dependencies]
bytecheck = "=0.8.2"
ptr_meta = { version = "=0.3.1", features = ["derive"] }
rancor = "=0.1.1"
rkyv = "=0.8.18"
```

`reproducer.rs`:

```rust
#![forbid(unsafe_code)]

//! Compile-only witness for rkyv 0.8.18.
//!
//! DO NOT execute `all_safe_ub_route`: reaching the indicated `rkyv::access`
//! call causes undefined behavior in the affected release. `main` deliberately
//! does not call it, so `cargo check` and a normal run of this file do not take
//! the UB path.

use core::alloc::{Layout, LayoutError};

// The CheckBytes derive expands to paths through this direct dependency.
use rancor as _;
use rkyv::{
    access,
    boxed::ArchivedBox,
    primitive::ArchivedUsize,
    rancor::Error,
    traits::{ArchivePointee, LayoutRaw},
};

#[derive(ptr_meta::Pointee, rkyv::Portable, bytecheck::CheckBytes)]
#[repr(C)]
struct Underreported {
    head: u8,
    tail: [u8],
}

// This implementation is accepted by a safe, public, unsealed trait. It lies
// about the complete layout of `Underreported` when the metadata is nonzero.
impl LayoutRaw for Underreported {
    fn layout_raw(_: usize) -> Result<Layout, LayoutError> {
        Layout::from_size_align(1, 1)
    }
}

impl ArchivePointee for Underreported {
    type ArchivedMetadata = ArchivedUsize;

    fn pointer_metadata(metadata: &Self::ArchivedMetadata) -> usize {
        metadata.to_native() as usize
    }
}

// rkyv's default archive format uses aligned little-endian 32-bit relative
// pointers. Alignment 4 makes the eight-byte root at bytes 4..12 aligned.
#[repr(align(4))]
struct Aligned([u8; 12]);

#[allow(dead_code)]
fn all_safe_ub_route() -> Result<(), Error> {
    // Bytes 4..8 encode relative offset -4; bytes 8..12 encode metadata 1024.
    // The ArchivedBox root therefore points to byte 0 as an Underreported DST
    // with a 1024-byte trailing slice. Only twelve bytes actually exist.
    let bytes = Aligned([
        0x00, 0x00, 0x00, 0x00, // target prefix
        0xfc, 0xff, 0xff, 0xff, // little-endian i32 -4
        0x00, 0x04, 0x00, 0x00, // little-endian u32 1024
    ]);

    let _: &ArchivedBox<Underreported> = access(&bytes.0)?;
    Ok(())
}

const _: () = assert!(core::mem::size_of::<ArchivedBox<Underreported>>() == 8);
const _: () = assert!(core::mem::align_of::<ArchivedBox<Underreported>>() == 4);
const _: () = assert!(core::mem::size_of::<Aligned>() == 12);
const _: () = assert!(core::mem::align_of::<Aligned>() == 4);

fn main() {}
```

No Rust Playground link is provided. The hosted Playground currently builds third-party crates with `default-features = false`, which removes rkyv's `bytecheck`-gated safe `access` route; it therefore cannot faithfully compile this configuration, and the service cannot accept the manifest needed to enable those features. The local Cargo manifest above enables rkyv's default features and pins the audited dependencies.

This local compile-check succeeded with stable Rust 1.93.1 on `x86_64-unknown-linux-gnu` against the exact pins above. The same source also compile-checked against current rkyv `main` at `fe0b6c04...`, and against a current compatible dependency resolution (`bytecheck 0.8.3`, `ptr_meta 0.3.2`, and `rancor 0.1.3`). Successful compilation establishes source selection, typing, coherence, and the compile-time size/alignment assertions; the UB claim comes from the source/dataflow proof above, not from executing the witness.

## Suggested remediation

The smallest direct repair appears to be:

1. Change `LayoutRaw` to `pub unsafe trait LayoutRaw`.
2. Add a precise `# Safety` contract requiring `layout_raw(m)` to return the exact complete size and alignment for `Self` with metadata `m`, or return `LayoutError` when that layout cannot be represented. State that this must hold for every metadata value accepted by consumers, not only metadata produced by the crate's serializers.
3. Change the four in-tree implementations (sized `T`, `[T]`, `str`, and `CStr`) to `unsafe impl` and add local proof comments for their exact layout computations.
4. Add a compile-fail UI test showing that the all-safe implementation above is rejected unless the implementer explicitly accepts an unsafe contract. Do not execute the UB path as a test.

Sealing `LayoutRaw` is another sound option if downstream custom DST implementations are not a supported extension point. Merely retaining the trait as safe and documenting the expected result does not close the soundness boundary, and checking the range against the same implementer-supplied layout is not an independent validation.

Making the trait unsafe is a small implementation diff, but it is a source-level compatibility change for downstream `LayoutRaw` implementations and should be called out in the release notes/security advisory.

## Related work and duplicate search

I found no exact public issue, pull request, or RustSec advisory for this safe-extension-trait defect as of 2026-09-04. Closed [#326](https://github.com/rkyv/rkyv/issues/326) and merged [#327](https://github.com/rkyv/rkyv/pull/327) are the closest history: they changed the built-in slice layout path from panicking on `Layout::array` failure to propagating `LayoutError`. They do not make `LayoutRaw` unsafe or sealed and do not prevent a custom safe implementation from returning an undersized `Ok` layout. [#670](https://github.com/rkyv/rkyv/issues/670) concerns shared-validation identity and DST metadata, while [#635](https://github.com/rkyv/rkyv/pull/635) concerns serialization output alignment; neither establishes this trait's layout proposition. The search is necessarily qualified because private reports and differently worded records may not be discoverable.

joshlf:
*Authored by an AI agent acting on joshlf's behalf.*
*This finding was identified using the [unsafe-rust skill](https://github.com/google/zerocopy/tree/main/skills/unsafe-rust).*

## Summary

`LayoutRaw` is consumed as proof of a pointee's complete allocation layout, but it was previously a safe, downstream-implementable trait. A safe implementation could underreport a dynamically sized pointee and let safe code perform out-of-allocation pointer arithmetic.

`LayoutRaw` is now an unsafe trait. Its contract requires every successful result to describe the exact complete layout of the pointee selected by the supplied metadata, including prefixes, padding, and a trailing dynamically sized field; implementations must return `LayoutError` when that exact layout cannot be represented.

This converts and documents all four active implementations (`Sized`, slice, `str`, and `CStr`) as audited `unsafe impl`s. A trybuild regression test also ensures that downstream implementations must explicitly cross the unsafe boundary. This prevents safe code from supplying an undersized layout that a safe allocation or access path later trusts for pointer arithmetic.

Fixes #692

## Testing

- `cargo +nightly fmt --all -- --check`
- `git diff --check`
- `cargo test --workspace --locked --offline` (164 `rkyv` unit tests, 6 derive integration tests, 7 fuzz regressions, and 82 doctests passed)
- `cargo +nightly test -p rkyv --test ui --locked --offline -- ui trybuild=layout_raw_requires_unsafe.rs`
- CI-style little-endian/aligned/pointer-width-32 tests and a big-endian/unaligned/pointer-width-64 `alloc + bytecheck` check
- `cargo +1.81.0 check -p rkyv --no-default-features --locked`
- workspace check, rustdoc, and `rkyv` library clippy checks with warnings denied

The complete installed-nightly UI run accepts the new case but has diagnostic drift in the pre-existing `enum_as` and `the_most_unhelpful_error` snapshots. The all-workspace/all-target clippy run likewise reaches only pre-existing lints in unrelated tests and examples; both targeted library clippy runs pass. The UB witness was not executed.

## Compatibility and scope

This is intentionally source-breaking for downstream `LayoutRaw` implementations: each must be audited and changed to `unsafe impl`. It does not change the archive format or built-in runtime behavior.

The commented-out `rkyv_dyn_derive` and `rkyv_dyn_test` crates remain dormant, stale, and unverified; their generated and handwritten implementations must be audited before those crates are re-enabled. The `CStr` proof is also version-scoped to `ptr_meta`'s byte-count metadata and the private transparent `CStr` representation observed from Rust 1.81 through the tested toolchain; a standard-library representation change is a re-audit trigger. This PR closes the `LayoutRaw` boundary issue only and is not a whole-crate soundness claim.


djkoloski:
Fixed by 1b052dc7d2e15aeb0abaa762ce1208aa640853e0

